PR #7, opened 2018-08-28
all: add initial support for Go modules

sbinet:
This CL was created by running:

$> GO111MODULE=on go mod init
$> GO111MODULE=on go mod tidy


teo:
I've rebased your PR on top of branch roletree (which is the current dev branch) but I believe the deps have changed compared to master. Could you please amend your commit?

teo:
@sbinet I've regenerated and pushed to development branch.
Thanks!

PR #39, opened 2019-05-10
Octrl 100/viper core

kostorr:


core/core.go
```diff
@@ -26,31 +26,29 @@ package core
...
 
 // Run is the entry point for this scheduler.
 // TODO: refactor Config to reflect our specific requirements
-func Run(cfg Config) error {
-	if cfg.veryVerbose {
-		cfg.verbose = true
-	}
-	if cfg.verbose {
+func Run() error {
+	if viper.GetBool("verbose") || viper.GetBool("veryVerbose") {
```
teo: Is it still the case that `veryVerbose true` also sets `verbose` to `true`?
Generally this `verbose` and `veryVerbose` stuff needs to be sorted out, unless we want to end up with `veryVeryVerbose` and `superDuperOmgWtfBbqVerboseAF` sometime down the line.
For now let's just make sure that `veryVerbose` implies `verbose` to preserve existing behavior.
For more information on future developments, see [OCTRL-68](https://alice.its.cern.ch/jira/browse/OCTRL-68)

core/core.go
```diff
@@ -26,31 +26,29 @@ package core
...
 		logrus.SetLevel(logrus.DebugLevel)
 	}
 
-	log.WithField("configuration", cfg).Debug("starting up")
+	log.WithField("configuration", viper.AllSettings()).Debug("starting up")
```
teo: I'd like to only do a full dump if `veryVerbose`.

core/scheduler.go
```diff
@@ -105,7 +106,7 @@ func runSchedulerController(ctx context.Context,
 	// ID, as well as additional information such as Roles, WebUI URL, etc.
 	return controller.Run(
 		ctx,
-		buildFrameworkInfo(state.config),
+		buildFrameworkInfo(),
```
teo: It is so satisfying to not have to pass `state.config` around any more.

core/state.go
```diff
@@ -29,43 +29,47 @@ package core
...
+	metricsAPI := initMetrics()
 	executorInfo, err := prepareExecutorInfo(
-		cfg.executor,
-		cfg.mesosExecutorImage,
-		buildWantsExecutorResources(cfg),
-		cfg.mesosJobRestartDelay,
+		viper.GetString("executor"),
+		viper.GetString("mesosExecutorImage"),
+		buildWantsExecutorResources(viper.GetFloat64("executorCPU"),
+			viper.GetFloat64("executorMemory")),
```
teo: This makes me think that in hindsight it might have been nice to have a strongly typed structure to unmarshal into, but we can leave it like this for the time being. Either way most of these values are only used in a few spots.

hacking/coreconfig.yaml
```diff
@@ -0,0 +1,36 @@
+---
```
teo: I think a file such as this one can be useful as documentation, so let's not delete it, but I think we should also have a config file which is a minimal, reasonable, not too verbose default configuration.
Proposal: rename this to `settings.yaml.example`, and put together a second `settings.yaml` which only overrides the defaults, if and where necessary. I'd expect this to boil down to just a few essential entries, including
```
controlport
executor
mesosurl
verbose
veryverbose
workflowconfigurationuri
```

core/config.go
```diff
@@ -25,133 +25,156 @@
...
+		return errors.New("Cannot find scheduler executable path: " + err.Error())
+	}
+	exeDir := filepath.Dir(exe)
+
+	viper.SetDefault("controlPort", 47102)
+	viper.SetDefault("coreConfigurationUri", "consul://127.0.0.1:8500") //TODO: TBD
+	viper.SetDefault("executor", env("EXEC_BINARY", filepath.Join(exeDir, "o2control-executor")))
+	viper.SetDefault("executorCPU", envFloat("EXEC_CPU", "0.01"))
+	viper.SetDefault("executorMemory", envFloat("EXEC_MEMORY", "64"))
+	viper.SetDefault("instanceName", "AliECS instance")
```
teo: This should default to `fmt.Sprintf("%s instance", product.PRETTY_SHORTNAME)`

PR #42, opened 2019-05-29
Make the run number a uint32

teo:
@graduta @awegrzyn this will be out in the next release and you'll need to pull the new `o2control.proto`.


graduta:
> @graduta @awegrzyn this will be out in the next release and you'll need to pull the new `o2control.proto`.

Hi, @teo . Was this released on aidrefflp01 and is good to go?

teo:
@graduta It's not released, but I can deploy master on aidrefflp01 if you like. I plan on releasing after I finish the InfoLogger refactor, ideally this week. Then you'll have a version number to rely on.

PR #47, opened 2019-07-02
OCTRL-82/git-managed-config

kostorr:


hacking/testings.yaml
```diff
@@ -1,6 +1,6 @@
 ---
```
kostorr: Will delete this before merging.

core/repos/repo.go
```diff
@@ -0,0 +1,210 @@
...
+
+	revSlice := strings.Split(repoPath, "@")
+
+	var repoUrlSlice []string
+	var revision string
+	if len(revSlice) == 2 { //revision specified
+		repoUrlSlice = strings.Split(revSlice[0], "/")
+		revision = revSlice[1]
+	} else if len(revSlice) == 1{ //no revision specified
+		repoUrlSlice = strings.Split(revSlice[0], "/")
```
teo: I'd advise validating the length of `repoUrlSlice` and returning error if `!= 3`, otherwise the `[]` in the return statement might fail.

core/repos/repo.go
```diff
@@ -0,0 +1,210 @@
...
+		repoUrlSlice[2], revision, false}, nil
+}
+
+func (r *Repo) GetIdentifier() string {
+	identifier := r.HostingSite + "/" + r.User + "/" + r.RepoName + "/"
+
+	return identifier
+}
+
+func (r *Repo) GetCompleteIdentifier() string { //unused
```
teo: Why do we have it if it's unused?

core/repos/repo.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+func (r *Repo) getCloneDir() string {
+	cloneDir := viper.GetString("repositoriesUri")
+	if cloneDir[len(cloneDir)-1:] != "/" {
+		cloneDir += "/"
+	}
+
+	cloneDir += r.HostingSite + "/" +
+				r.User 		 + "/" +
```
teo: Whitespace issue?

core/repos/repo.go
```diff
@@ -0,0 +1,210 @@
...
+		r.HostingSite + "/" +
+		r.User 		  + "/" +
+		r.RepoName	  + ".git"
+}
+
+func (r *Repo) GetTaskDir() string {
+	return r.getCloneDir() + "/tasks/"
+}
+
+func (r *Repo) getWorkflowDir() string {
```
teo: Is there a reason for `GetTaskDir` public but `getWorkflowDir` private?

core/repos/repomanager.go
```diff
@@ -0,0 +1,389 @@
...
+	"os"
+	"sort"
+	"strings"
+	"sync"
+)
+
+var (
+	once sync.Once
+	instance *RepoManager
+	gitAuthUser = "kalexopo"
```
teo: Is a single auth mechanism/credentials used for all the repos? I figure it should be per-repo.
I also expect it shouldn't require additional configuration storage, as once the repo is cloned the credentials should be stored in `.git/config` (except with password auth, but we can choose not to support this).

core/repos/repomanager.go
```diff
@@ -0,0 +1,389 @@
...
+
+func Instance() *RepoManager {
+	once.Do(func() {
+		instance = initializeRepos()
+	})
+	return instance
+}
+
+type RepoManager struct {
+	repoList map[string]*Repo //I want this to work as a set
```
teo: Is this still supposed to be a set? Shouldn't it be `map[*Repo]bool` then? Or are you now satisfied with the structure as it is?

core/repos/repomanager.go
```diff
@@ -0,0 +1,389 @@
...
+
+		_, err = git.PlainClone(repo.getCloneDir(), false, &git.CloneOptions{
+			Auth:   auth,
+			URL:    repo.getUrl(),
+			ReferenceName: plumbing.NewBranchReferenceName(repo.Revision),
+		})
+
+
+		if err != nil {
+			if err.Error() == "repository already exists" { //Make sure master is checked out
```
teo: Can we make this check more reliable?

core/repos/repomanager.go
```diff
@@ -0,0 +1,389 @@
...
+			err = os.Remove(dir)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func isEmpty(path string) (bool, error) { //TODO: Teo, should this be put somewhere like utils?
```
teo: Fine with me. I advise a new package `utils` in `common`.

core/repos/repomanager.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	return nil
+}
+
+func (manager *RepoManager) RefreshRepo(repoPath string) error {
+	manager.mutex.Lock()
+	defer manager.mutex.Unlock()
+
+	if !strings.HasSuffix(repoPath, "/") { //Add trailing '/'
```
teo: You seem to do this often. Fine as long as it works, but I'd suggest a tiny utility function for ensuring the trailing slash is there. Compilers these days are good enough to inline such situations.

core/repos/repomanager.go
```diff
@@ -0,0 +1,389 @@
...
+	} else if newDefaultRepo == manager.defaultRepo {
+		return errors.New(newDefaultRepo.GetIdentifier() + " is already the default repo")
+	}
+
+	manager.setDefaultRepo(newDefaultRepo)
+
+	return nil
+}
+
+func (manager *RepoManager) EnsureReposPresent(taskClassesRequired []string) (err error) {
```
teo: I don't understand what this does. Do we automatically add and clone repos for cached task classes even if these repos aren't already in repo manager? Is such a starting condition even possible?

core/state.go
```diff
@@ -109,10 +109,10 @@ func newInternalState(shutdown func()) (*internalState, error) {
 
 	taskman := task.NewManager(resourceOffersDone,
 		tasksToDeploy, reviveOffersTrg, state.commandqueue)
-	err = taskman.RefreshClasses()
+	/*err = taskman.RefreshClasses()
```
teo: This is commented out, not a problem, but why is it so?

core/task/class.go
```diff
@@ -41,6 +42,7 @@ type TaskClass info
 //   role info.
 type info struct {
 	Name        string                  `yaml:"name"`
```
teo: Is the Name field still used for other things besides unmarshaling?

core/task/class.go
```diff
@@ -51,6 +53,19 @@ type info struct {
...
 }
 
+type taskClassIdentifier struct {
+	repo repos.Repo
+	Name string
+}
+
+func (tcID taskClassIdentifier) String() string {
+	if tcID.repo.Revision != "" {
+		return fmt.Sprintf("%vtasks/%v@%v", tcID.repo.GetIdentifier(), tcID.Name, tcID.repo.Revision)
```
teo: `%s` should work here if all substitution inputs are strings.

core/task/manager.go
```diff
@@ -98,28 +102,92 @@ func (m*Manager) NewTaskForMesosOffer(offer *mesos.Offer, descriptor *Descriptor
 	return
 }
 
-func (m *Manager) RefreshClasses() (err error) {
+// TODO: Any reason to be a *Manager method? If not then should it reside here?
```
teo: It doesn't have to be in `task.Manager`, it can also be in `repo.Manager` if we can make it independent from the `task` package.

core/task/manager.go
```diff
@@ -98,28 +102,92 @@ func (m*Manager) NewTaskForMesosOffer(offer *mesos.Offer, descriptor *Descriptor
...
+		repo = repoManager.GetRepos()[repo.GetIdentifier()] //get repo pointer from repomanager
+	 	if repo == nil { //should never end up here
+			return nil, errors.New("getTaskClassList: repo not found for " + taskClass)
+		}
+
+		yamlData, err = ioutil.ReadFile(viper.GetString("repositoriesUri") + taskClassFile)
+		if err != nil {
+			return nil, err
+		}
+		//var taskClass *TaskClass //TODO: This doesn't unmarshal; unclear why
```
teo: Let's figure this out together.

core/workflow/aggregatorrole.go
```diff
@@ -76,15 +76,16 @@ func (r *aggregatorRole) GlobFilter(g glob.Glob) (rs []Role) {
 	return
 }
 
-func (r *aggregatorRole) ProcessTemplates() (err error) {
+
+func (r *aggregatorRole) ProcessTemplates(repo *repos.Repo) (err error) {
```
teo: For clarity I'd always call this argument to the various implementations of `ProcessTemplates` `workflowRepo` instead of just `repo`.

core/repos/repo.go
kostorr: Remove this.

core/config.go
```diff
@@ -118,6 +122,7 @@ func setFlags() error {
 	pflag.Int("metrics.port", viper.GetInt("metrics.port"), "Port of metrics server (listens on server.address)")
 	pflag.String("metrics.path", viper.GetString("metrics.path"), "URI path to metrics endpoint")
 	pflag.Bool("summaryMetrics", viper.GetBool("summaryMetrics"), "Collect summary metrics for tasks launched per-offer-cycle, offer processing time, etc.")
+	pflag.String("repositoriesUri", viper.GetString("repositoriesUri"), "URI path to git-managed configuration repositories")
 	pflag.Bool("verbose", viper.GetBool("verbose"), "Verbose logging")
 	pflag.Bool("veryVerbose", viper.GetBool("veryVerbose"), "Very verbose logging")
 	pflag.String("workflowConfigurationUri", viper.GetString("workflowConfigurationUri"), "URI of the Consul server or YAML configuration file, used for workflow configuration.")
```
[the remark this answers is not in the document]
kostorr: It's there in case we want to point to a different directory than the default. If this is not foreseen it should be removed, yes.

core/repos/repomanager.go
kostorr: Yes, this ensures that when a workflow asks for tasks that come from a different repo, said repo is present (if not, it is added) and that the expected revision is checked out.

Task classes are no longer cached, they are pulled on demand, whenever they are requested by a workflow. After that they will remain cached, until a RefreshClasses() call is made, which will remove only if they are inactive.

core/task/class.go
kostorr: Nope, moved the unmarshalling directly to taskIdentifier.

core/task/class.go
kostorr: Done

core/workflow/aggregatorrole.go
kostorr: Done

core/task/manager.go
kostorr: Can't be independent from the `task` package and thus creates an import cycle. Will leave it here.

core/repos/repomanager.go
kostorr: Comment was leftover.

core/task/manager.go
kostorr: No luck with this one.

teo: Now I know why: if you look at a task class file, for example https://gitlab.cern.ch/tmrnjava/aliecs_conf/blob/master/tasks/readout.yaml you can notice that there's a `-` in front of everything. Translating that to JSON, the file contains a structure like `[{'name':'readout', ...}]` rather than `{'name':'readout', ...}`.
Since we expect to have one and only one task class descriptor per task class file, the file should be modified as follows, so that the top level node is the dict itself and not an array:
```
name: readout # read by workflow
control:
  mode: direct
wants:
  cpu: 0.15
  memory: 128
bind: []
properties: {}
command: 
  env: []
  shell: true
  arguments:
    - "file:/opt/alisw/el7/Readout/v0.24-occ-1/etc/readout.cfg"
  value: >-
    source /etc/profile.d/modules.sh &&
    MODULEPATH=/opt/alisw/el7/modulefiles module load Readout &&
    MODULEPATH=/opt/alisw/el7/modulefiles module load Control-OCCPlugin &&
    MODULEPATH=/opt/alisw/el7/modulefiles module list &&
    readout.exe
```
Then you'll be able to unmarshal into the *TaskClass directly instead of the slice.

teo:
A few more issues I've noticed while testing:
* If the `repositoriesPath` isn't writable, AliECS core should complain immediately and not only after the first invocation to `coconut repo`.
* `workflowConfigurationUri` should be renamed to `globalConfigurationUri` or something else, as workflows don't live there any more.
* `coconut template list` should query for available workflow templates in all the repos (@master), not the old way.
* `FATAL create:      command finished with error error=rpc error: code = Internal desc = cannot create new environment: cannot load workflow template: open ~/.aliecs.d/reposgitlab.cern.ch/kalexopo/AliECS_conf/tasks/readout.yaml: no such file or directory` -- it should be ok to put in a repos path both with and without trailing `/`.
* The setting of default repo is lost between core restarts.

core/task/manager.go
teo: I've taken the liberty of porting the sample tasks and workflows from `config.yaml` to this repo https://gitlab.cern.ch/tmrnjava/aliecs_conf so feel free to use them when testing this. Generally once we can load the FairMQ sample workflows, we can trust everything is in place to merge.

coconut/control/control.go
```diff
@@ -504,9 +505,165 @@ func ListWorkflowTemplates(cxt context.Context, rpc *coconut.RpcClient, cmd *cob
 		fmt.Fprintln(o, "no templates found")
 	} else {
 		fmt.Fprintln(o, "available templates in loaded configuration:")
+		var prevRepo string
 		for _, tmpl := range templates {
-			fmt.Fprintln(o, "\t" + tmpl)
+			if (prevRepo != tmpl.GetRepo()) {
```
sbinet:
```suggestion
			if prevRepo != tmpl.GetRepo() {
```

coconut/control/control.go
```diff
@@ -504,9 +505,165 @@ func ListWorkflowTemplates(cxt context.Context, rpc *coconut.RpcClient, cmd *cob
...
 		}
 	}
 	return nil
+}
+
+// ListRepos lists all available git repositories that are used for configuration.
+func ListRepos(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.Command, args []string, o io.Writer) (err error) {
+	if len(args) != 0 {
+		err = errors.New(fmt.Sprintf("accepts no args, received %d", len(args)))
+		return
```
sbinet: naked returns are frown upon (and source of subtle bugs).

```suggestion
		return err
```

coconut/control/control.go
```diff
@@ -504,9 +505,165 @@ func ListWorkflowTemplates(cxt context.Context, rpc *coconut.RpcClient, cmd *cob
...
+	}
+
+	var response *pb.ListReposReply
+	response, err = rpc.ListRepos(cxt, &pb.ListReposRequest{}, grpc.EmptyCallOption{})
+	if err != nil {
+		return
```
sbinet: ditto.
```suggestion
		return err
```

coconut/control/control.go
```diff
@@ -504,9 +505,165 @@ func ListWorkflowTemplates(cxt context.Context, rpc *coconut.RpcClient, cmd *cob
...
+			if root.GetDefault() {
+				defaultTick = blue("YES")
+			}
+			table.Append([]string{strconv.Itoa(i), root.GetName(), defaultTick})
+		}
+		fmt.Fprintf(o, "Git repositories used as configuration sources:\n\n")
+		table.Render()
+	}
+
+	return
```
sbinet:
```suggestion
	return err
```

coconut/control/control.go
```diff
@@ -504,9 +505,165 @@ func ListWorkflowTemplates(cxt context.Context, rpc *coconut.RpcClient, cmd *cob
...
+}
+
+// AddRepo add a new repository to the available git repositories used for configuration and checks it out.
+func AddRepo(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.Command, args []string, o io.Writer) (err error) {
+	if len(args) != 1 {
+		err = errors.New(fmt.Sprintf("accepts 1 arg, received %d", len(args)))
+		return
```
sbinet:
```suggestion
		return err
```

coconut/control/control.go
```diff
@@ -504,9 +505,165 @@ func ListWorkflowTemplates(cxt context.Context, rpc *coconut.RpcClient, cmd *cob
...
+	}
+
+	var response *pb.AddRepoReply
+	response, err = rpc.AddRepo(cxt, &pb.AddRepoRequest{Name: args[0]}, grpc.EmptyCallOption{})
+	if err != nil {
+		return
```
sbinet:
```suggestion
		return err
```

coconut/control/control.go
```diff
@@ -504,9 +505,165 @@ func ListWorkflowTemplates(cxt context.Context, rpc *coconut.RpcClient, cmd *cob
...
+	}
+
+	if response.GetErrorString() == "" {
+		fmt.Fprintln(o, "Repository succesfully added.")
+	} else {
+		fmt.Fprintln(o, "Cannot add repository:", response.GetErrorString())
+	}
+
+	return
```
sbinet:
```suggestion
	return err
```

core/confsys/service.go
```diff
@@ -100,6 +100,29 @@ func newService(uri string) (svc *Service, err error) {
 	return &Service{src: src}, err
 }
 
+func (s *Service) NewDefaultRepo(defaultRepo string) (err error) {
```
sbinet:
```suggestion
func (s *Service) NewDefaultRepo(defaultRepo string) error {
```

kostorr: Thanks!

PR #67, opened 2019-09-17
OCTRL-138

kostorr:


core/protos/o2control.proto
```diff
@@ -276,11 +276,17 @@ message GetRolesReply {
     repeated RoleInfo roles = 1;
 }
 
-message GetWorkflowTemplatesRequest{}
+message GetWorkflowTemplatesRequest{
```
graduta: Will the request still work with no parameters? Right now ControlGUI makes use of this method and is not providing any parameters

coconut/cmd/template_list.go
```diff
@@ -35,10 +35,31 @@ var templateListCmd = &cobra.Command{
 	Aliases: []string{"list", "ls", "l"},
 	Short: "list available workflow templates",
 	Long: `The template list command shows a list of available workflow templates.
-These workflow templates can then be loaded to create an environment.`,
+These workflow templates can then be loaded to create an environment.
```
graduta: Examples can be added as a separate field in cobra and Long only kept as a sentence describing the use. 
```
Example: `coconut conf list
coconut conf list <component>
coconut conf list <component> -t <timestamp>`,
```

core/protos/o2control.proto
kostorr: No, you can pass the following for the same behaviour.

```
repoPattern="",
revisionPattern="",
allBranches=false,
allTags=false 
```

kostorr: I could always add a wrapper of course...

coconut/cmd/template_list.go
kostorr: Thanks for the tip!

core/protos/o2control.proto
graduta: Right now we are not passing anything so as long as the request will default those values to empty is fine by me

coconut/control/control.go
```diff
@@ -490,13 +490,89 @@ func QueryRoles(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.Command,
...
+				fmt.Fprintln(o, "Ignoring `--all-{branches,tags}` flags, as a valid revision has been specified")
+				allBranches = false
+				allTags = false
+			} else  {
+				revisionPattern = "*"
+			}
+		}
+
+	} else 	if len(args) == 1 { // If we have an argument, give priority over the flags
+		slicedArgument := strings.Split(args[0], "@")
```
graduta: This might be safer with a regex just in case the user passes repo-patter@

teo: I agree that a regex would be the most elegant way to make a comprehensive but early check. The present example is likely low-impact though, as we already make sure that `len(slicedArgument)` is 1 or 2.

core/protos/o2control.proto
teo: In `proto3` all fields are optional, and in memory a zero value is treated by Protobuf the same as an undefined field.
Thus, I would expect the following query
```
repoPattern="",
revisionPattern="",
allBranches=false,
allTags=false 
```
to be pushed by Protobuf as a message with all its fields `null`, and I'd expect the core to return a list of templates from all repos, at `HEAD` of the `master` branch.

core/protos/o2control.proto
```diff
@@ -276,11 +276,17 @@ message GetRolesReply {
...
+    string revisionPattern = 2;
+    bool allBranches = 3;
+    bool allTags = 4;
+}
 
 message WorkflowTemplateInfo {
     string repo = 1;
-    string template = 2;
+    string revision = 2;
+    string template = 3;
```
teo: Can we do this without reindexing fields?
```
string template = 2;
string revision = 3;
```
should work just as well, and it would improve `proto` interface compatibility.

core/repos/repo.go
```diff
@@ -187,16 +188,80 @@ func (r *Repo) refresh() error {
...
 	if err != nil {
-		return workflows, err
+		return nil, err
 	}
-	for _, file := range files {
-		if strings.HasSuffix(file.Name(), ".yaml") { // Only return .yaml files
-			workflows = append(workflows, strings.TrimSuffix(file.Name(), ".yaml"))
+
+	workflows := make(map[string][]string)
+	for _, revision := range revisionsMatched {
```
teo: I'd appreciate a few words of comment on what we're about to do here.

core/repos/repo.go
```diff
@@ -187,16 +188,80 @@ func (r *Repo) refresh() error {
...
+
+	// Check if the revision pattern is actually a hash. If so return a single revision
+	hashMaybe := plumbing.NewHash(revisionPattern)
+	resolvedHash, err := ref.ResolveRevision(plumbing.Revision(revisionPattern))
+	if err == nil && *resolvedHash == hashMaybe {
+		revisions := append(revisions, resolvedHash.String())
+		return revisions, nil
+	}
+
+	err = refs.ForEach(func(ref *plumbing.Reference) error {
```
teo: I'd appreciate a few words of comment on what's about to happen here.

core/repos/repomanager.go
```diff
@@ -192,11 +194,33 @@ func cleanCloneParentDirs(parentDirs []string) error {
 	return nil
 }
 
-func (manager *RepoManager) GetRepos() (repoList map[string]*Repo) {
+func (manager *RepoManager) GetRepos(repoPattern ...string) (repoList map[string]*Repo) { // Simulate function overloading; for 0 or 1 arguments here
```
teo: Proposal:
```
func (manager *RepoManager) GetAllRepos() (repoList map[string]*Repo) 
func (manager *RepoManager) GetRepos(repoPattern string) (repoList map[string]*Repo) 
```

core/repos/repomanager.go
```diff
@@ -192,11 +194,33 @@ func cleanCloneParentDirs(parentDirs []string) error {
...
+	for _, repo := range manager.repoList {
+		if g.Match(repo.GetIdentifier()) {
+			matchedRepoList[repo.GetIdentifier()] = repo
+		}
+	}
+
+	return matchedRepoList
+}
+
+func (manager *RepoManager) getRepoByIndex(index int) (*Repo, error) {
```
teo: TODO: deduplicate functionality from repo remove/default/refresh.

core/repos/repomanager.go
```diff
@@ -408,18 +427,60 @@ func (manager *RepoManager) GetOrderedRepolistKeys() []string {
...
+		}
 
 	}
 
 	return templateList, numTemplates, nil
-}
\ No newline at end of file
+}
+
+const (
```
teo: Goes on top

coconut/cmd/template_list.go
```diff
@@ -35,10 +35,30 @@ var templateListCmd = &cobra.Command{
...
+ * ` + "`coconut templ list --revision=5c7f1c1fded1b87243998579ed876c8035a08377 `" + ` lists templates from the commit corresponding to the hash for all git repositories`,
+
 	Run:   control.WrapCall(control.ListWorkflowTemplates),
 }
 
 func init() {
 	templateCmd.AddCommand(templateListCmd)
-}
+
+	templateListCmd.Flags().String("repo", "*", "repositories to list templates from")
```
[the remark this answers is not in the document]
kostorr: There is no alias for flags, only a shorthand which can be a character long. I rename to repository and didn't add the 'r' shorthand as it may be confused with the revision.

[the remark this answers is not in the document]
teo: Let's make it `--repository` and `-p` as shorthand, and let's add the `p` shorthand to all other situations where the word `repository` is a flag or a subcommand.

coconut/control/control.go
```diff
@@ -490,13 +491,89 @@ func QueryRoles(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.Command,
...
+		if match, _ := regexp.MatchString("\\A[^@]+\\z", args[0]); match {
```
teo: Rather than ignoring the error, I'd rather have `regexp.MustCompile` on the expression, and then https://golang.org/pkg/regexp/#Regexp.MatchString instead of https://golang.org/pkg/regexp/#MatchString

PR #70, opened 2019-09-27
[OCONF-184] coconut conf history command

graduta:


coconut/cmd/configuration_history.go
```diff
@@ -0,0 +1,48 @@
+/*
+ * === This file is part of ALICE O² ===
+ *
+ * Copyright 2018 CERN and copyright holders of ALICE O².
```
teo: Please make sure that the year is correct in all headers.

coconut/cmd/configuration_history.go
```diff
@@ -0,0 +1,48 @@
...
+
+package cmd
+
+import (
+	"github.com/AliceO2Group/Control/coconut/configuration"
+	"github.com/spf13/cobra"
+)
+
+var configurationHistoryCmd = &cobra.Command{
+	Use:   "history [component]",
```
teo: I'd use angled <> brackets here, as the component is mandatory. I'd also call it "query" as it can include just the component, or also the entry.

coconut/cmd/configuration_history.go
```diff
@@ -0,0 +1,48 @@
...
+	Aliases: []string{"h"},
+	Example: `coconut conf history <component>
+coconut conf history <component> <entry>
+coconut conf history <component>/<entry>`,
+	Short: "List all existing entries with timestamps of a specified component in Consul",
+	Long: `The configuration history command will return all entries with
```
teo: s/will return/returns/

coconut/configuration/configuration.go
```diff
@@ -64,6 +67,15 @@ const  (
 	logicError = iota // Logic/Output error
 )
 
+var(
+	white = color.New(color.FgHiWhite).SprintFunc()
+	blue = color.New(color.FgHiBlue).SprintFunc()
+	green = color.New(color.FgHiGreen).SprintFunc()
+	yellow = color.New(color.FgHiYellow).SprintFunc()
+	red = color.New(color.FgHiRed).SprintFunc()
+	grey = color.New(color.FgWhite).SprintFunc()
```
teo: Do we use all of these? Are they duplicated with the control utils?

coconut/configuration/configuration.go
```diff
@@ -262,6 +274,68 @@ func Show(cfg *configuration.ConsulSource, cmd *cobra.Command, args []string, o
 	return nil, nonZero
 }
 
+func History(cfg *configuration.ConsulSource, cmd *cobra.Command, args []string, o io.Writer)(err error, code int) {
+	var key, component, entry string
+
+	if len(args) < 1 ||  len(args) > 2 {
+		err = errors.New(fmt.Sprintf(" accepts between 0 and 3 arg(s), but received %d", len(args)))
```
teo: Is the leading space deliberate?

coconut/configuration/configuration.go
```diff
@@ -343,4 +427,42 @@ func GetListOfComponentsAndOrWithTimestamps(keys []string, keyPrefix string, use
...
+
+func GetComponentEntryTimestamp(key string)(string, string, string) {
+	key = strings.TrimPrefix(key, componentsPath)
+	key = strings.TrimPrefix(key, "/'")
+	key = strings.TrimSuffix(key, "/")
+	elements := strings.Split(key, "/")
+	return elements[0], elements[1], elements[2]
+}
+
+func GetMaxLenOfKey(keys []string) (maxLen int){
```
teo: Are these kind of functions ever used outside this configuration package? If not, they should be private.

graduta: The plan is to move them to configuration_util. Similar to what control_util is.
Also I need them to be exported for testing purposes. Thus, I made them exportable from start

coconut/configuration/configuration.go
graduta: No, it is not. Updated accordingly

coconut/configuration/configuration.go
graduta: Removed unused ones

coconut/cmd/configuration_history.go
graduta: 👍

coconut/cmd/configuration_history.go
graduta: 👍

coconut/cmd/configuration_history.go
graduta: 👍

PR #71, opened 2019-10-01
[OCONF-166] coconut conf import command

graduta:


teo:
Fixes please:
## import command
* s/Extension of the file should be/Supported file extensions. To force a specific configuration format, see flag --format/-f/
* s/Please check through the already existing components:/Available components in configuration database:/
* In all simple lists (starting with `-`), make sure that there's a single space between `-` and the first letter of the list item.
* s/If you wish to add a new component, please use flag `-n/-new-component` to create a new component/To create a new component, see flag --new-component/-n/
* s/A new component has been created/New component created/    s/A new entry has been created/New entry created/   s/The following configuration key has been added/Configuration key created/
* s/Component `read-teo-test` already exists, thus flag  `-n/-new-component` cannot be used/Invalid use of flag --new-component/-n: component red(read-teo-test) already exists/
* s/Component read-teo-test has been updated//
* s/Entry myEntry2 has been updated/Entry updated: myEntry2/
## show command
* s/The configuration show command requests by default the latest 
configuration for the specified component and entry. It can request exact 
time configuration by specifying wanted timestamp as flag/The configuration show command returns the most recent configuration revision for the specified component and entry. It can also return a specific revision, requested with the --timestamp/-t flag./

Suggestions:
* Look into empty lines at the end of output and remove them.

coconut/cmd/configuration_import.go
```diff
@@ -0,0 +1,55 @@
...
+	Use:   "import <component> <entry> <file_path>",
+	Aliases: []string{"i", "imp"},
+	Example: `coconut conf import <component> <entry> <file_path>
+coconut conf import <component/entry> <file_path>
+coconut conf import <component> <entry> <file_path> -n
+coconut conf import <component/entry> <file_path> --new-component
+coconut conf import <component/entry> <file_path> --format=json
+coconut conf import <component/entry> <file_path>.json
+`,
+	Short: "Import a configuration file for the component and entry specified",
```
teo: `Import a configuration file for the specified component and entry`

coconut/cmd/configuration_import.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"github.com/AliceO2Group/Control/coconut/configuration"
+	"github.com/spf13/cobra"
+)
+
+var configurationImportCmd = &cobra.Command{
+	Use:   "import <component> <entry> <file_path>",
+	Aliases: []string{"i", "imp"},
+	Example: `coconut conf import <component> <entry> <file_path>
+coconut conf import <component/entry> <file_path>
```
teo: Minor nitpick: I'd put `<component>/<entry>` instead of `<component/entry>` as `component` and `entry` are the variables here, which we compose in one way or another.

coconut/cmd/configuration_import.go
```diff
@@ -0,0 +1,55 @@
...
+coconut conf import <component> <entry> <file_path> -n
+coconut conf import <component/entry> <file_path> --new-component
+coconut conf import <component/entry> <file_path> --format=json
+coconut conf import <component/entry> <file_path>.json
+`,
+	Short: "Import a configuration file for the component and entry specified",
+	Long: `The configuration import command will generate a timestamp and save
```
teo: s/will generate/generates/

coconut/cmd/configuration_import.go
```diff
@@ -0,0 +1,55 @@
...
+the configuration file under the component/entry/timestamp path in Consul. Default
```
teo: `... save the configuration file to Consul under the path <component>/<entry>/<timestamp>.`

coconut/cmd/configuration_import.go
```diff
@@ -0,0 +1,55 @@
...
+accepted file extensions are JSON, YAML, TOML, INI`,
```
teo: s/Default accepted file extensions are/Supported configuration file types are JSON, YAML, TOML and INI, and their file extensions are recognized automatically./

coconut/cmd/configuration_import.go
```diff
@@ -0,0 +1,55 @@
...
+	Run: configuration.WrapCall(configuration.Import),
+	Args:  cobra.ExactArgs(3),
+}
+
+func init() {
+	configurationCmd.AddCommand(configurationImportCmd)
+	configurationImportCmd.Flags().BoolP("new-component", "n",  false, "used to add a new component")
+	configurationImportCmd.Flags().StringP("format", "f",  "", "used to specify a different extension file than default (JSON, YAML, INI, TOML)")
```
teo: `force a specific configuration file type, overriding any file extension`

coconut/cmd/configuration_import.go
```diff
@@ -0,0 +1,55 @@
...
+	Long: `The configuration import command will generate a timestamp and save
+the configuration file under the component/entry/timestamp path in Consul. Default
+accepted file extensions are JSON, YAML, TOML, INI`,
+	Run: configuration.WrapCall(configuration.Import),
+	Args:  cobra.ExactArgs(3),
+}
+
+func init() {
+	configurationCmd.AddCommand(configurationImportCmd)
+	configurationImportCmd.Flags().BoolP("new-component", "n",  false, "used to add a new component")
```
teo: `create a new configuration component while importing entry`

coconut/configuration/configuration.go
```diff
@@ -332,133 +315,94 @@ func History(cfg *configuration.ConsulSource, cmd *cobra.Command, args []string,
...
-		} else {
-			componentsFullName = componentParts[0]
+	components := getComponentsMapFromKeysList(keys)
+	componentExist := components[component]
+	if !componentExist &&  !useNewComponent {
+		componentMsg := ""
+		for key, _ := range components {
+			componentMsg += "\n-" + key
 		}
+		return errors.New(fmt.Sprintf("Component `" + component + "` does not exist. " +
```
teo: Let's not use backticks here, either color or nothing. The backticks can potentially make sense in Cobra command doc strings because those are exported as Markdown, but not in error messages.

coconut/configuration/configurationutil.go
```diff
@@ -0,0 +1,224 @@
...
+		if parts[0] == component {
+			entriesMap[parts[1]] = true
+		}
+	}
+	return entriesMap
+}
+
+func isFileExtensionValid(extension string) bool{
+	extension = strings.ToUpper(extension)
+	return extension == "JSON" || extension == "YAML" || extension == "INI" || extension == "TOML"
```
teo: Let's also allow `.yml` just in case. Visible strings should not be changed, just the logic to approve the file extension.

coconut/cmd/configuration_import.go
graduta: 👍

coconut/configuration/configuration.go
```diff
@@ -155,34 +147,29 @@ func List(cfg *configuration.ConsulSource, cmd *cobra.Command, args []string, o
...
-			err = errors.New(fmt.Sprintf("Flag `-t / --timestamp` could not be identified"))
-			return err, invalidArgs
+			return err,  invalidArgs
 		}
 		if len(args) == 1 {
-			if !IsInputSingleValidWord(args[0]) {
-				err = errors.New(fmt.Sprintf("Requested component name cannot contain character `/` or `@`"))
-				return err, invalidArgs
+			if !isInputSingleValidWord(args[0]) {
+				return  errors.New(fmt.Sprintf(invalidArgsErrMsg)), invalidArgs
```
kostorr: This may simply be `errors.New(invalidArgsErrMsg)`. [https://golang.org/pkg/errors/#New](https://golang.org/pkg/errors/#New)
`fmt.Sprintf() ` is not necessary if you don't have some fancy arguments you want to handle.

graduta: Did not know. Thank you 👍

coconut/configuration/configuration.go
```diff
@@ -332,133 +315,94 @@ func History(cfg *configuration.ConsulSource, cmd *cobra.Command, args []string,
...
+	}
+	if componentExist && useNewComponent {
+		return errors.New(fmt.Sprintf("Component `" + component + "` already exists, thus flag  `-n/-new-component` cannot be used" )), invalidArgs
+	}
 
-		if strings.Compare(componentsSet[componentsFullName], componentTimestamp) < 0{
-			componentsSet[componentsFullName] = componentTimestamp
-		}
+	entryExists := false
+	if useNewComponent {
```
kostorr: Is this clause necessary?

coconut/configuration/configurationutil.go
```diff
@@ -0,0 +1,224 @@
...
+func getLatestTimestamp(keys []string, component string, entry string)(timestamp string, err error, code int) {
+	keyPrefix := componentsPath + component + "/" + entry
+	if len(keys) == 0 {
+		err = errors.New(fmt.Sprintf("No keys found"))
+		return "", err, emptyData
+	}
+
+	var maxTimeStamp uint64
+	for _, key := range keys {
+		componentTimestamp, err := strconv.ParseUint(strings.TrimPrefix(key, keyPrefix + "/"), 10, 64)
```
kostorr: I imagine this is done on purpose here, but this leads to variable shadowing, which could in turn lead to errors. A very good explanation can be found [here](https://rpeshkov.net/blog/golang-variable-shadowing/).

coconut/configuration/configuration.go
graduta: Not anymore. I negated it

coconut/configuration/configurationutil.go
graduta: In this case I am not interested in stopping the whole command if one of the timestamps is bad that is why I am adding it to the list of elements only if it is a valid one. So even if the timestamp is not correct, it will be ignored.

The error that I am interested here is the one with emptyData

coconut/configuration/configuration.go
teo: There's also `fmt.Errorf` that combines the two for when you **do** need to handle fancy arguments ;)

coconut/cmd/configuration_import.go
```diff
@@ -33,23 +33,22 @@ var configurationImportCmd = &cobra.Command{
...
 	Aliases: []string{"i", "imp"},
 	Example: `coconut conf import <component> <entry> <file_path>
-coconut conf import <component/entry> <file_path>
-coconut conf import <component> <entry> <file_path> -n
-coconut conf import <component/entry> <file_path> --new-component
-coconut conf import <component/entry> <file_path> --format=json
-coconut conf import <component/entry> <file_path>.json
+coconut conf import <component> <entry> <file_path> --new-component
+coconut conf import <component> <entry> <file_path> --format=json
+coconut conf import <component> <entry> <file_path>.json
```
teo: We seem to be missing the `<component>/<entry>` bits.

coconut/configuration/configuration.go
```diff
@@ -55,11 +55,10 @@ type ConfigurationCall func(*configuration.ConsulSource, *cobra.Command, []strin
 const  (
 	nonZero = iota
 	invalidArgs = iota // Provided args by the user are invalid
-	invalidArgsErrMsg = "Component and Entry names cannot contain `/ or  `@`"
+	invalidArgsErrMsg = "component and Entry names cannot contain `/ or  `@`"
```
teo: s/Entry/entry/

coconut/configuration/configuration.go
```diff
@@ -390,17 +387,15 @@ func Import(cfg *configuration.ConsulSource, cmd *cobra.Command, args []string,
...
 	if !entryExists {
-		userMsg += "A new entry has been created: " + blue(entry) + "\n"
+		userMsg += "New entry created: " + blue(entry) + "\n"
 	} else {
-		userMsg += "Entry " + blue(entry) + " has been updated" +  "\n"
+		userMsg += "Entry updated: " + blue(entry) +  "\n"
 	}
 	fullKey :=  red(component) + "/" + blue(entry) + "@" + strconv.FormatInt(timestamp, 10)
-	userMsg += "The following configuration key has been added: " + fullKey
+	userMsg += "Configuration created: " + fullKey
```
teo: I know we discussed this at length trying to find the best wording and we settled on this, but after additional thought I realized that:
1) this string is the only one that's always printed (the ones for new entry and component aren't)
2) this string is the last (and sometimes only) line of output in response to a `coconut conf import` command.
---
Therefore I propose the following:
`userMsg += "Configuration imported: " + fullKey`

coconut/cmd/configuration_import.go
graduta: They are not missing as the requirement was to have exactly 3 arguments:
```
coconut conf import <component name> <entry name> <file path> [options]
```

teo: I see, so `coconut conf import <component/entry> <file_path>.json` never worked even when it was documented?

graduta: No. It was a mistake when documentation was added and on second look of the requirements it was then removed. 

If we wish to support that we can added so that we are in line with the 
```coconut conf show command```

teo: Yes, that would be splendid, and also the final thing before we merge.

teo: The feature implementation looks good, could you please just add back the relevant documentation lines for the `<component>/<entry>` syntax? Once this is done we can merge.

PR #73, opened 2019-10-18
WIP: [core] Update repo rpcs to return errors

kostorr:
_protos not generated_


coconut/control/control.go
```diff
@@ -710,28 +702,23 @@ func RefreshRepos(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.Comman
...
 
-	errorString := response.GetErrorString()
-	if errorString == "" {
-		if len(args) == 0 {
-			fmt.Fprintln(o, "Repositories refreshed succesfully")
-		} else {
-			fmt.Fprintln(o, "Repository refreshed succesfully")
-		}
+	if len(args) == 0 {
+		fmt.Fprintln(o, "Repositories refreshed succesfully")
```
graduta: Both if & else condition are doing the same thing. Is the If condition still needed?

kostorr: One is "repositories" and the other is "repository".

PR #83, opened 2020-02-07
[OCTRL-73] Properties Tasks

graduta:
Changes to add properties from a task-class defined in ControlWorkflows to running tasks of an environment


teo:
Merged into kvstore in #83

PR #85, opened 2020-02-07
[core+coconut] Support a configurable default revision for workflow repos

kostorr:
Close OCTRL-166


coconut/control/control.go
```diff
@@ -622,39 +622,63 @@ func ListRepos(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.Command,
...
 
 	return nil
 }
 
 // AddRepo add a new repository to the available git repositories used for configuration and checks it out.
 func AddRepo(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.Command, args []string, o io.Writer) (err error) {
-	if len(args) != 1 {
-		err = errors.New(fmt.Sprintf("accepts 1 arg, received %d", len(args)))
+
+	var name string
```
graduta: Optional

```
var name, defaultRevision string
```
or 
``` 
names, defaultRevision :="", ""
```

coconut/cmd/repo_default_revision.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/AliceO2Group/Control/coconut/control"
+	"github.com/spf13/cobra"
+)
+
+// repoDefaultRevision sets global and per-repo default revision
+var repoDefaultRevisionCmd = &cobra.Command{
+	Use:   "default-revision",
+	Short: "set default global and per-repository revision",
+	Long: fmt.Sprintf(`The repository default-revision command sets the global default
```
graduta: I can see that in some commands we use 
```
Long : fmt.Sprintf(`Some text`)
```
while in others we do
```
Long: `Some text`
```

@teo is there any standard or recommended way?

teo: @graduta it really depends on what you need to do. Generally we should only wrap in `Sprintf` if we're formatting a string: if you don't have `%s` and `%d` and such, `Sprintf` should be omitted.

The use of plain quotes vs backticks is simply convenience, backticks are a verbatim environment, they make sense for long strings that include line breaks.

coconut/control/control.go
kostorr: Thanks!

coconut/cmd/repo_default_revision.go
kostorr: I changed all repo commands that unnecessarily used `fmt.Sprintf` to "strings".

core/repos/repomanager.go
```diff
@@ -134,21 +145,81 @@ func (manager *RepoManager)  discoverRepos() (repos []string, err error){
...
+		ReferenceName: plumbing.NewBranchReferenceName("master"), //Check out master so we have a working copy
+	})
+
+	if err != nil && err != git.ErrRepositoryAlreadyExists {
+			return err
+	}
+
+	// Decide if revision = defaultRevision -> if yes lock them together
+	// We do this because, in the case where the revision was not specified, it would fall back to the default revision
+	// As a result, subsequent changes to the default revision should be followed by revision
```
teo: This is not how I understood it, but it might be a good idea. Let's discuss.

core/repos/repomanager.go
```diff
@@ -134,21 +145,81 @@ func (manager *RepoManager)  discoverRepos() (repos []string, err error){
...
+	// Check that the defaultRevision is valid
+	var prefixes []string
+	prefixes = append(prefixes, refRemotePrefix, refTagPrefix)
+	var matchedRevs []string
+	matchedRevs, err = repo.getRevisions(repo.DefaultRevision, prefixes)
+	if err != nil {
+		return err
+	} else if len(matchedRevs) == 0 {
+		log.Warning("Default revision " + repo.DefaultRevision + " invalid for " + repo.GetIdentifier())
+		if repo.DefaultRevision != manager.defaultRevision { //TODO: This check implies that repo.DefaultRevision might be different to the global on. This is not the case yet but it stays here for reference.
```
teo: Not sure I understand the TODO. You mean that it will be possibly different from the global default once we load persistent per-repo defaults?

PR #87, opened 2020-02-07
[core]:Persistent FID

miltalex:
* Create a single writable directory path. 
* Persistent Mesos Framework ID.
    * Store Mesos Framework ID to Consul under `o2/aliecs` or to a local file under `/var/lib/o2/aliecs`.
   * Create reconciliation call on every re-registration.
   * Kill tasks received from reconciliation.

Close OCTRL-170

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -27,6 +27,7 @@ package core
 import (
 	"context"
 	"github.com/AliceO2Group/Control/common/product"
+	"github.com/AliceO2Group/Control/core/the"
 	"github.com/spf13/viper"
 	"time"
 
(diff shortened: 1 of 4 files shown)


core/core.go
```diff
@@ -72,13 +73,23 @@ func Run() error {
...
 	fidStore := store.DecorateSingleton(
 		store.NewInMemorySingleton(),
 		store.DoSet().AndThen(func(_ store.Setter, v string, _ error) error {
-			log.WithField("frameworkId", v).Debug("generated new frameworkId")
+			// Store Mesos Framework ID to configuration.
+			err = the.ConfSvc().NewMesosFID(v)
+			if err != nil {
+				log.WithField("error", err).Error("cannot write to configuration")
+			}
+			log.WithField("frameworkId", v).Debug("frameworkId")
```
teo: Am I right in assuming that the `AndThen` here runs after every `Set` of the `fidStore`, the first occasion of which happens immediately after subscription when an empty `fidStore` receives a FID from Mesos?

miltalex: That's correct.

PR #88, opened 2020-02-10
[core]: repos and runcounter.txt are stored inside coreWorkingDir

miltalex:
Closes OCTRL-129.

### DIFF
--- a/common/utils/utils.go
+++ b/common/utils/utils.go
@@ -53,3 +53,9 @@ func EnsureTrailingSlash(path *string) {
 
 // helper func to package strings up nicely for protobuf
 func ProtoString(s string) *string { return &s }
+
+func RemoveTrailingSlash(path *string) {
+	if strings.HasSuffix(*path, "/") { //Remove trailing '/'
+		strings.TrimRight(*path, "/")
+	}
+}
--- a/common/utils/utils.go
+++ b/common/utils/utils.go
@@ -53,9 +53,3 @@ func EnsureTrailingSlash(path *string) {
 
 // helper func to package strings up nicely for protobuf
 func ProtoString(s string) *string { return &s }
-
-func RemoveTrailingSlash(path *string) {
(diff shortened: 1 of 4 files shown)


core/confsys/service.go
```diff
@@ -129,19 +125,8 @@ func (s *Service) NewRunNumber() (runNumber uint32, err error) {
...
 		var rnf string
-		rnf, err = s.src.Get("o2/control/run_number_file")
-		if err != nil {
-			return
-		}
-		rnf, err = homedir.Expand(rnf) // Resolve ~ into home dir
-		if err != nil {
-			return
-		}
+		rnf = viper.GetString("coreWorkingDir") + "runcounter.txt"
```
teo: Are we certain that `coreWorkingDir` has a trailing slash?
Generally it shouldn't: the default shouldn't have a trailing `/`, and if the user passes a `/path/with/trailing/slash/`, it should be sanitized to remove it.

Then, this line becomes
```
		rnf = viper.GetString("coreWorkingDir") + "/runcounter.txt"
```

teo: Addendum: all user-provided paths must be sanitized to ensure the absence of a trailing slash.

miltalex: The `coreWorkingDir` is sanitized [here](https://github.com/AliceO2Group/Control/blob/master/core/config.go#L184).  I followed the same pattern used for the `repositoriesPath`.

common/utils/utils.go
```diff
@@ -53,3 +53,9 @@ func EnsureTrailingSlash(path *string) {
 
 // helper func to package strings up nicely for protobuf
 func ProtoString(s string) *string { return &s }
+
+func RemoveTrailingSlash(path *string) {
```
teo: To be checked if there is an `os.path` function that can already do this for us.

miltalex: Change it to `filepath.Clean()`

PR #90, opened 2020-02-17
[core]: remove repositoriesPath from configuration

miltalex:
Closes OCTRL-184.

### DIFF
--- a/core/confsys/service.go
+++ b/core/confsys/service.go
@@ -103,7 +103,7 @@ func (s *Service) NewDefaultRepo(defaultRepo string) error {
 		return cSrc.Put("o2/control/default_repo", defaultRepo)
 	} else {
 		data := []byte(defaultRepo)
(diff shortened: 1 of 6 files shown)


core/confsys/service.go
```diff
@@ -103,7 +103,7 @@ func (s *Service) NewDefaultRepo(defaultRepo string) error {
 		return cSrc.Put("o2/control/default_repo", defaultRepo)
 	} else {
 		data := []byte(defaultRepo)
-		return ioutil.WriteFile(viper.GetString("repositoriesPath") + "default_repo", data, 0644)
+		return ioutil.WriteFile(filepath.Join(viper.GetString("coreWorkingDir"), "repos", "default_repo"), data, 0644)
```
teo: Looks like we do a lot of this `filepath.Join` stuff all over `service.go` and the `repos` package.

Could you please make a `func (s *Service) GetReposPath() string` which always returns the correctly formatted repos directory path based on the current value of `viper.GetString("coreWorkingDir")`?

PR #94, opened 2020-03-06
[OCTRL-206] implement stop destroy option

graduta:
Add an option to DestroyEnvironment option to destroy it even when in running state if the user specifies so. 

* Added flag to coconut
* Added parameter to GRPC API Call


coconut/control/control.go
```diff
@@ -420,26 +420,17 @@ func DestroyEnvironment(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.
...
+	keepTasks, err := cmd.Flags().GetBool("keep-tasks")
 	if err != nil {
-		return
-	}
-
-	allowedState := "CONFIGURED"
-	if envResponse.GetEnvironment().GetState() != allowedState {
-		fmt.Fprint(o, "cannot teardown environment\n")
-		fmt.Fprintf(o, "teardown is allowed in state %s, but environment %s is in state %s\n", allowedState, envId, envResponse.GetEnvironment().GetState())
-		return
```
teo: How does this affect the quality of the error text provided by `coconut` when the environment cannot be torn down due to a disallowed state?

core/server.go
```diff
@@ -299,8 +299,16 @@ func (m *RpcServer) DestroyEnvironment(cxt context.Context, req *pb.DestroyEnvir
 		return nil, status.Newf(codes.NotFound, "environment not found: %s", err.Error()).Err()
 	}
 
-	statesForDestroy := [...]string{"CONFIGURED", "STANDBY"}
+	if req.AllowInRunningState && env.CurrentState() == "RUNNING" {
+		err = env.TryTransition(environment.MakeTransition(m.state.taskman, pb.ControlEnvironmentRequest_CONFIGURE))
```
teo: This should be a `ControlEnvironmentRequest_STOP_ACTIVITY`, otherwise the transition will necessarily fail.

graduta: 👍

coconut/control/control.go
graduta: `DestroyEnvironment` from server.go will send back with an error with a similar message if the environment cannot be destroyed due to the current state:
""cannot destroy environment in state %s"

PR #97, opened 2020-03-18
WIP [OCTRL-209] Add no-version to configuration import

graduta:
Add 'no-versioning' option to `coconut configuration import` command.

Cases to treat:
* Using `no-versioning`:
  * Entry specified already exists and contains timestamps => Return error and inform users
  * 
* With versioning:
  * Entry specified already exists and contains no-versioning items => Return error and inform users`

### DIFF
--- a/coconut/cmd/configuration_import.go
+++ b/coconut/cmd/configuration_import.go
@@ -37,8 +37,8 @@ coconut conf import <component>/<entry> <file_path>
 coconut conf import <component> <entry> <file_path> --new-component
 coconut conf import <component>/<entry> <file_path> --format=json
 coconut conf import <component> <entry> <file_path>.json
-coconut conf import <component> <entry> <file_path> -no-versioning 
-coconut conf import <component> <entry> <file_path> -no-versioning --new-component
+coconut conf import <component> <entry> <file_path> --no-versioning 
+coconut conf import <component> <entry> <file_path> --no-versioning --new-component
 `,
 	Short: "Import a configuration file for the specified component and entry",
 	Long: `The configuration import command generates a timestamp and saves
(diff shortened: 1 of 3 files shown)


coconut/cmd/configuration_import.go
```diff
@@ -37,6 +37,8 @@ coconut conf import <component>/<entry> <file_path>
 coconut conf import <component> <entry> <file_path> --new-component
 coconut conf import <component>/<entry> <file_path> --format=json
 coconut conf import <component> <entry> <file_path>.json
+coconut conf import <component> <entry> <file_path> -no-versioning
```
teo: `--no-versioning` double dash

coconut/cmd/configuration_import.go
```diff
@@ -51,5 +53,5 @@ func init() {
 	configurationCmd.AddCommand(configurationImportCmd)
 	configurationImportCmd.Flags().BoolP("new-component", "n",  false, "create a new configuration component while importing entry")
 	configurationImportCmd.Flags().StringP("format", "f",  "", "force a specific configuration file type, overriding any file extension")
-
+	configurationImportCmd.Flags().Bool("no-versioning",  false, "create a new configuration component without timestamp being generated as a version")
```
teo: `"create an unversioned configuration entry (no timestamps are stored)"`

coconut/configuration/configuration.go
```diff
@@ -391,9 +397,35 @@ func Import(cfg *configuration.ConsulSource, cmd *cobra.Command, args []string,
...
+			// If a timestamp already exists in the entry specified by the user, than it cannot be used
+			return errors.New("Specified entry: '" + entry + "' already contains versioned items. Please " +
+				"specify a different entry name"), invalidArgs
+		}
+		if latestTimestamp == "0" && !useNoVersion {
+			// If a timestamp does not exist for specified entry but user wants versioning than an error is thrown
+			return errors.New("Specified entry: '" + entry + "' already contains un-versioned items. Please " +
+				"specify a different entry name"), invalidArgs
+		}
+	}
```
teo: Yes very good this is exactly what I had in mind. Versioned and unversioned entries we simply don't mix.

PR #98, opened 2020-03-18
[OCTRL-209] Support unversioned entries for rest of coconut conf commands

graduta:


coconut/cmd/configuration_list.go
```diff
@@ -34,7 +34,7 @@ var configurationListCmd = &cobra.Command{
 	Aliases: []string{"l", "ls"},
 	Example: `coconut conf list
 coconut conf list <component>
-coconut conf list <component> -t <timestamp>`,
+coconut conf list <component> -t`,
```
teo: Good catch!

PR #100, opened 2020-03-20
[coconut] Fix repo default-revision bug & improve documentation

kostorr:
Close OCTRL-211

### DIFF
--- a/coconut/cmd/repo_default_revision.go
+++ b/coconut/cmd/repo_default_revision.go
@@ -25,6 +25,7 @@
 package cmd
 
 import (
+	"fmt"
 	"github.com/AliceO2Group/Control/coconut/control"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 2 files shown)


coconut/cmd/repo_default_revision.go
```diff
@@ -33,7 +34,14 @@ import (
 var repoDefaultRevisionCmd = &cobra.Command{
 	Use:   "default-revision",
 	Short: "set default global and per-repository revision",
-	Long: "The repository default-revision command sets the global default repository revision as well as the per-repository default revision.",
+	Long: fmt.Sprintf(`The repository default-revision command sets the global default repository revision.
```
graduta: No need for `fmt.Sprintf` here

teo: That's true, as you're using `+` to concatenate rather than a format string. No big deal either way.

kostorr: Indeed, I thought that it was necessary for the line changes. I'll push a quick fix.

PR #112, opened 2020-04-24
Handle external kill of controlled process

miltalex:
Closes OCTRL-148, OCTRL-223.


core/environment/manager.go
```diff
@@ -115,6 +115,33 @@ func (envs *Manager) CreateEnvironment(workflowPath string, userVars map[string]
...
+		env.wfAdapter.SubscribeToStateChange(subscriptionId, notify)
+		defer env.wfAdapter.UnsubscribeFromStateChange(subscriptionId)
+
+		wfState := wf.GetState()
+		if wfState != task.ERROR {
+			WORKFLOW_STATE_LOOP:
+			for {
+				select {
+				case wfState = <-notify:
+					if wfState == task.ERROR {
```
teo, line 135: I suspect we should also listen for `task.DONE`, or for some other event on environment destruction so this goroutine dies with the environment.

core/environment/environment.go
```diff
@@ -208,4 +208,15 @@ func (env *Environment) GetCurrentRunNumber() (rn uint32) {
...
 	return env.currentRunNumber
+}
+
+// setState will move environment to a given state from current state.
+// The call does not trigger any callbacks, if defined.
+func (env *Environment) setState(state string) {
+	if env == nil {
+		return
+	}
+	env.Mu.RLock()
```
teo, line 219: Since we're writing to the `Sm`, I suspect we should use `Lock` instead of `RLock` here.

PR #114, opened 2020-05-06
Allow templating and overriding of channel configuration

miltalex:
Closes OCTRL-214.

### DIFF
--- a/core/task/descriptor.go
+++ b/core/task/descriptor.go
@@ -29,14 +29,14 @@ import (
 )
 
 /*
-1) Role has a method GenerateTaskConfiguration []Descriptor.
+1) Role has a method GenerateTaskDescriptors []Descriptor.
    For most roles, this function simply consolidates the output of itself for children.
    For TaskRoles, this function generates a single Descriptor from its own Class
    and parent role vars.
    Descriptor must also have a Ptr or path to the TaskRole that should become the new
    Task's parent.
    So in ConfigureTransition.do, we first get the []Descriptor with
-   wf.GenerateTaskConfiguration().
+   wf.GenerateTaskDescriptors().
 2) Now that we have a list of config items for tasks to run, we can port
    task.Manager.AcquireRoles. This one internally uses match.go facilities such as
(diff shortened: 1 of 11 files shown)


core/task/descriptor.go
```diff
@@ -26,17 +26,18 @@ package task
 
 import (
 	"github.com/AliceO2Group/Control/core/task/constraint"
+	"github.com/AliceO2Group/Control/core/task/channel"
 )
 
 /*
-1) Role has a method GenerateTaskConfiguration []Descriptor.
+1) Role has a method GenerateTaskDescriptors []Descriptor.
```
teo, line 33: Good catch

core/task/descriptor.go
```diff
@@ -48,6 +49,8 @@ type Descriptor struct {
 	TaskRole          parentRole
 	TaskClassName     string
 	RoleConstraints   constraint.Constraints
+	RoleConnect       []channel.Outbound
+	RoleInbound       []channel.Inbound
```
teo, line 53: Do we need these 2 in the `Descriptor` because we need to be able to access the full role hierarchy to do a merge from `scheduler.go`?

core/workflow/rolebase.go
```diff
@@ -153,14 +154,9 @@ func (r *roleBase) CollectOutboundChannels() (channels []channel.Outbound) {
 	if r.parent == nil {
 		channels = make([]channel.Outbound, 0)
 	} else {
-		channels = r.parent.CollectOutboundChannels()
-	}
-	for _, v := range r.Connect {
-		channels = append(channels, v)
-		// FIXME: this does not take into account child roles with outbound channels with the same name
-		// as an outbound channel in the parent.
-		// The correct behaviour would be OVERRIDE, currently the behavior is UNDEFINED.
```
teo, line 162: Glad to be rid of this FIXME

core/task/descriptor.go
miltalex: Yeah exactly.

PR #120, opened 2020-05-14
[core] create GetConsulPath for confsys

miltalex:
Closes OCTRL-195.


core/config.go
```diff
@@ -51,6 +51,7 @@ func setDefaults() error {
 
 	viper.SetDefault("controlPort", 47102)
 	viper.SetDefault("coreConfigurationUri", "consul://127.0.0.1:8500") //TODO: TBD
+	viper.SetDefault("consulBasePath", "o2/control")
```
teo, line 54: Would anything break on upgrade if we defaulted to `o2/aliecs` instead of `o2/control`?

miltalex: No, it should work since most values are being populated during deployment. The only problem will be that values that are created during runtime such as `run_number` ( will start from one instead of using any previous values stored under `o2/control`.

teo: @miltalex Ok, then please let's default to `o2/aliecs` and we merge.

PR #121, opened 2020-05-25
[walnut] Init walnut with basic schema validation

obviyus:


walnut/app/schemata.go
```diff
@@ -0,0 +1,245 @@
...
+                            "title": "Receive Buffer Size",
+                            "type": "number"
+                        },
+                        "rateLogging": {
+                            "title": "Rate Logging",
+                            "type": "number"
+                        }
+                    }
+                },
+                "task": {
```
teo, line 225: If possible, a check should ensure that in a role block either `task` exists, or `roles`, but not both.

`load` is required in `task`.

walnut/cmd/check.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+// checkCmd represents the check command
+var checkCmd = &cobra.Command{
+	Use:   "check",
+	Short: "check the file passed against a specified schema.",
+	Long: `The check command validates the given file against a specified 
+schema. This file can be a task template, a workflow template or an O2
```
teo, line 41: This is a minor nitpick, but for consistency I advise you use "O²". All Go strings are UTF-8 anyway.

walnut/cmd/check.go
```diff
@@ -0,0 +1,69 @@
...
+// checkCmd represents the check command
+var checkCmd = &cobra.Command{
+	Use:   "check",
+	Short: "check the file passed against a specified schema.",
+	Long: `The check command validates the given file against a specified 
+schema. This file can be a task template, a workflow template or an O2 
+DPL Dump. Each of those have a schema provided to validate against. 
+
+Usage:
+  walnut check --format [template] [file]
```
teo, line 45: I would not call it `template` here, it is bound to cause confusion since the actual files are called task templates and workflow templates.
Suggestion: `walnut check --format [format] [file]`

walnut/cmd/check.go
```diff
@@ -0,0 +1,69 @@
...
+
+Example:
+  walnut check --format workflow_template readout-sftb.yaml
+
+Valid schemas:
```
teo, line 50: I'm ok with both `schemas` and `schemata`, and so are dictionaries from what I can tell. But we should pick only one plural form and stick with it for consistency. Your other file is `schemata.go`, so it stands to reason that we should say "Valid schemata" here.

walnut/cmd/check.go
```diff
@@ -0,0 +1,69 @@
...
+  workflow  task  dpl_dump`,
+
+	Run: func(cmd *cobra.Command, args []string) {
+		fmt.Println("check called with arg: " + strings.Join(args, " "))
```
teo, line 54: It seems that this file, as well as `root.go`, are still a bit work in progress so I'll refrain from going into too much detail.

walnut/cmd/check.go
```diff
@@ -0,0 +1,69 @@
...
+			validate.Template(filename, format)
+		}
+	},
+	// Args: cobra.ExactArgs(1),
+}
+
+func init() {
+	rootCmd.AddCommand(checkCmd)
+
+	checkCmd.Flags().StringP("format", "f", "", "format to validate against")
```
teo, line 67: Very good that you added the shorthand :100:

walnut/cmd/root.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func init() {
+	cobra.OnInitialize(initConfig)
+
+	// Here you will define your flags and configuration settings.
+	// Cobra supports persistent flags, which, if defined here,
+	// will be global for your application.
+
+	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.walnut.yaml)")
```
teo, line 68: This is a good approach for storing `walnut` configuration, but the path is less than ideal on Linux these days.
I propose `$HOME/.config/walnut/settings.yaml`. See `coconut/cmd/root.go` for an example of this.

walnut/validate/validate.go
```diff
@@ -0,0 +1,100 @@
...
+	"os"
+
+	"github.com/AliceO2Group/Control/walnut/app"
+
+	"github.com/xeipuuv/gojsonschema"
+	"gopkg.in/yaml.v2"
+)
+
+// Template accepts a filename and format then validate against the schema specified (either workflow or task)
+func Template(filename string, format string) {
```
teo, line 39: I suggest we call it `CheckSchema` or `Check` or `SyntaxCheck` or even `Validate` (your pick).
`Template` has a specific meaning, and there is no template processing (i.e. resolution of `{{  }}` blocks) going on here, just pure YAML validation.

walnut/validate/validate.go
```diff
@@ -0,0 +1,100 @@
...
+	} else {
+		err := "Schema validation failed."
+		fmt.Printf("ERROR: %v", err)
+		os.Exit(1)
+	}
+}
+
+// convert takes a interface{} as input and recursively converts all child
+// map[interface{}]interface{} to map[string]interface{}
+func convert(i interface{}) interface{} {
```
teo, line 86: This approach is ok.

I suspect the difficulty in unmarshaling to `map[string]interface{}` derives from the fact that YAML allows complex non-string keys. An alternative approach to yours would be to define a `type map[string]interface{} MyMapType` and implement `func (m *MyMapType) UnmarshalYAML(unmarshal func(interface{}) error) (err error)`. By handling the string keys in this function you make it possible to unmarshal directly to `MyMapType`. Try if you like, or we can leave it as it is (I don't expect a performance difference in our use case).

walnut/validate/validate.go
obviyus: I had avoided using `Validate`,  as calling this function would result in `validate.Validate()`. I think `CheckSchema` works well with `validate.CheckSchema()`. Although I would've preferred `validate.Schema()` but that's a bit repetitive.

walnut/app/schemata.go
```diff
@@ -0,0 +1,245 @@
...
+ * In applying this license CERN does not waive the privileges and
+ * immunities granted to it by virtue of its status as an
+ * Intergovernmental Organization or submit itself to any jurisdiction.
+ */
+
+package app
+
+// TaskSchema is the task template schema, i.e. the expected format of
+// the task templates necessary for AliECS.
+var TaskSchema string = `
```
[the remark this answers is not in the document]
obviyus: I decided to go with `go generate`. Let me know if the current implementation satisfied you. I think there's still some scope for improvement.

walnut/validate/validate.go
obviyus: I'm not sure I completely understand this approach. It does seem like a cleaner solution, let me know if you think it's worth exploring. I'm fine with it either way.

walnut/app/schemata.go
```diff
@@ -0,0 +1,245 @@
...
+        "name": {
+            "description": "The name of the root role, which happens to be an aggregator role.",
+            "type": "string"
+        },
+        "defaults": {
+            "type": "object",
+            "title": "Variable definitions",
+            "description": "Variable definitions: defaults, overridden by anything",
+            "properties": {
+                "nProcessors": {
```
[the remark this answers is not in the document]
obviyus: Let me know if the changes are ok.

walnut/app/schemata.go
```diff
@@ -0,0 +1,245 @@
...
+        "roles": {
+            "title": "Roles",
+            "description": "A list of child roles.",
+            "type": "array",
+            "properties": {
+                "name": {
+                    "description": "Parametrized name of an iterator role",
+                    "type": "string"
+                },
+                "for": {
```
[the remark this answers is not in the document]
obviyus: I've used `#ref` to set role as a definition. This way it can have nested `roles`. Let me know if this approach is okay.

walnut/validate/validate.go
teo: `validate.CheckSchema()` is ok for me.

If you prefer something shorter, you could also consider renaming the `validate` package to `schema`, which gets you `schema.Validate()`.

walnut/validate/validate.go
teo: It is not critical in this specific case, but I expect understanding and using the `yaml.Unmarshaler`/`yaml.Marshaler` interfaces is stuff you'll need for the workflow conversion effort that follows, where non-trivial YAML scenarios make your current approach impractical.
I advise checking out https://godoc.org/gopkg.in/yaml.v2 as well as the files in https://github.com/AliceO2Group/Control/tree/master/core/workflow and https://github.com/AliceO2Group/Control/tree/master/core/task which use `yaml.Unmarshaler` extensively to control YAML-to-memory deserialization.

For example, if we consider `channel.go`, implementing the `UnmarshalYAML` function allows us to deserialize those `sndBufSize`, `rateLogging` etc. into an auxiliary type (`_channel`) which directly mimics the YAML structure. We then do some casts and fill in some defaults, and store this in the actual target object whose type doesn't directly match YAML without this special constructor.

walnut/schemata/includeSchemata.go
```diff
@@ -0,0 +1,63 @@
+/*
```
teo, line 1: I'd prefer if we kept file names all lowercase, so `includeschemata.go`.

walnut/schemata/includeSchemata_test.go
```diff
@@ -0,0 +1,33 @@
+/*
```
teo, line 1: Appreciate the test. Should be renamed to all lowercase.

walnut/validate/validate.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+// CheckSchema accepts a filename and format then validate against the schema specified (either workflow or task)
+func CheckSchema(filename string, format string) {
+
+	rawYAML, err := ioutil.ReadFile(filename) // import YAML file
+	if err != nil {
+		fmt.Printf("ReadFile failed: %v", err)
+		os.Exit(1)
+	}
```
teo, line 45: I'd probably keep the file reading outside `CheckSchema` and pass the `rawYAML` as an argument instead of the filename. This way any errors coming from `CheckSchema` are related to the actual check and not to basic file opening.

walnut/validate/validate.go
```diff
@@ -0,0 +1,100 @@
...
+
+	var dataFromYAML interface{}
+	if err := yaml.Unmarshal([]byte(rawYAML), &dataFromYAML); err != nil {
+		fmt.Printf("Unmarshaling YAML failed: %v", err)
```
teo, line 49: Instead of `fmt.Printf` here, I advise returning the same message in a wrapped error, i.e. `return fmt.Errorf("Unmarshaling YAML failed: %w", err)`. The caller should then check the error, print as needed and `os.Exit`.

See https://blog.golang.org/go1.13-errors for more information on error wrapping.

A function like `CheckSchema` should never `os.Exit(1)`, in command line utilities one has to be careful because it's very easy to lose track. It should return `error` to the caller, which should check whether the `error` is `nil` and act accordingly.

You generally want your `os.Exit` calls to be few and in places where you control exactly what and how you print out the error and the exit code.

walnut/validate/validate.go
```diff
@@ -0,0 +1,100 @@
...
+	dataFromYAML = convert(dataFromYAML)
+
+	var schema string
+	switch format {
+	case "task":
+		schema = schemata.Task
+	case "workflow":
+		schema = schemata.Workflow
+	default:
+		fmt.Print("ERROR: Invalid format specified.")
```
teo, line 62: Error messages should start lowercase, and they should usually not start with `ERROR:` as this is redundant. We know it's an error because we put it in an object of type `error`, which makes it different from a plain string.

walnut/validate/validate.go
obviyus: How about merging the two packages `schemata` and `validate` into just `schemata`? I don't think we need the `validate` function to be separate from the package that generates the schema.

walnut/app/schemata.go
obviyus: Added in 1cc187a

walnut/validate/validate.go
obviyus: Understood.

walnut/validate/validate.go
teo: Sure, fine with me.

PR #128, opened 2020-06-08
[WIP] [walnut] Loading DPL dumps into workflow.Role

obviyus:
An importer tool to load DPL Dumps generated by `O2Suite/latest` into a `workflow.Role` struct.


walnut/converter/converter.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+
+		var task = task.Class{
+			Identifier: task.TaskClassIdentifier{
+				Name: DPL.Workflows[index].Name,
+			},
+			Defaults: nil,
+			Control: struct {
+				Mode controlmode.ControlMode "yaml:\"mode\""
+			}{Mode: 0},
```
obviyus, line 77: Not sure how to handle this. `ControlMode` is of type `int`. How do I pass `fairmq`?

walnut/converter/dump.json
```diff
@@ -0,0 +1,364 @@
...
+                    "description": "RAWDATA",
+                    "subspec": 0,
+                    "lifetime": 0
+                }
+            ],
+            "options": [],
+            "rank": 0,
+            "nSlots": 1,
+            "inputTimeSliceId": 0,
+            "maxInputTimeslices": 1
```
obviyus, line 31: Not sure if it's a mistake but `maxInputTimeSlices` would be a better name, don't you think?

walnut/converter/converter.go
teo: You should be able to pass the constant `controlmode.FAIRMQ` from https://github.com/AliceO2Group/Control/blob/c8f974dc0f0819369210243e46139830bb5e48d6/common/controlmode/controlmode.go

walnut/converter/dump.json
teo: It comes from DPL. Your suggestion might indeed be more consistent...
Pinging @ktf in case he'd like to fix it... or confirm whether it's formatted like this on purpose.

core/task/channel/channel.go
```diff
@@ -25,33 +25,33 @@
...
-	RcvBufSize  int                     `yaml:"rcvBufSize"`
-	RateLogging int                     `yaml:"rateLogging"`
-	Transport   TransportType           `yaml:"transport"`  //default: default
+type Channel struct {
+	Name        string        `yaml:"name"`
+	Type        ChannelType   `yaml:"type"`
+	SndBufSize  int           `yaml:"sndBufSize"`
+	RcvBufSize  int           `yaml:"rcvBufSize"`
+	RateLogging int           `yaml:"rateLogging"`
+	Transport   TransportType `yaml:"transport"` //default: default
```
teo, line 44: No need to change back these files, but I suggest you disable formatting on save in your IDE. I like nicely formatted code but commits should generally only touch what they need to touch, otherwise it becomes difficult later on to understand which part of the commit is a functional change as opposed to formatting.

walnut/converter/importer.go
```diff
@@ -0,0 +1,102 @@
...
+type Dump struct {
+	Workflows []workflow `json:"workflow"`
+	Metadatas []metadata `json:"metadata"`
+}
+
+func jsonImporter(input *os.File) (importedJSON Dump, err error) {
+	scanner := bufio.NewScanner(input)
+	var inputJSON string
+
+	// TODO: Probably a better way to handle this
```
teo, line 76: Not really...
@ktf is there a way to ensure we only get the workflow dump and nothing else? As things are now we usually get a bunch of log messages before the JSON starts.

walnut/converter/dump.json
ktf: Nah, just a typo I guess. PR welcome. ;-)

walnut/converter/converter.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/AliceO2Group/Control/core/task/constraint"
+	"gopkg.in/yaml.v3"
+)
+
+// return pointer to float64
+func create(x float64) *float64 {
+	return &x
+}
+
+type class struct {
```
obviyus, line 45: I had to duplicate this definition to get `MarshalYAML` on L134 to work. Is there a way to achieve the same result while also avoiding this?

walnut/converter/dump.json
```diff
@@ -0,0 +1,364 @@
...
+                {
+                    "binding": "out",
+                    "origin": "TST",
+                    "description": "RAWDATA",
+                    "subspec": 0,
+                    "lifetime": 0
+                }
+            ],
+            "options": [],
+            "rank": 0,
```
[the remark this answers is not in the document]
ktf: There is nothing in DPL at the moment stating how much memory / cpu a given workflow will use. Does that really matter? In any case placement will not happen based on the resources available but on the functionality, no?

walnut/converter/dump.json
```diff
@@ -0,0 +1,364 @@
...
+            "nSlots": 1,
+            "inputTimeSliceId": 0,
+            "maxInputTimeslices": 1
+        },
+        {
+            "name": "Dispatcher",
+            "inputs": [
```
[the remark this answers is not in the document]
ktf: The correspondence is calculated by DPL, based on the actual InputSpec and OutputSpec. The channel names are indeed `from_{sender}_to_{receiver}-{extra}`. Bind vs connect or the channel type is define by a policy. For the latter two I guess you can do whatever you think it's best for your setup. For the channel name, I guess I will have to expose those?

walnut/converter/importer.go
ktf: https://alice.its.cern.ch/jira/browse/O2-1513

walnut/converter/dump.json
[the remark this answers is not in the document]
obviyus: I'm not sure about the policy but from what I can guess:
- `bind` is only ever used for tasks
- `connect` is only used for workflows
- `bind` forms a `push` channel by default
- `connect` forms a `pull` channel by default

Let me know if these sound correct to you. 

Exposing the channel names would be great. I have not been able to infer what the `{extra}` field should be from the DPL Dumps.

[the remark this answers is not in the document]
ktf: regarding bind / connect at the moment we have that sources always bind and sinks always connect, but that's just how it is setup by default.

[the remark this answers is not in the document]
ktf: same for push / pull (actually now I use a pair, IIRC...) source always push, sink always pull, but again, in certain cases they might be customised.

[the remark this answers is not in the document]
teo: @ktf How would it be reflected in the DPL dump if they were customized?

walnut/converter/dump.json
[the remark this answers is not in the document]
teo: @ktf Correct. If you envisage for the DPL user to define this kind of metadata we'd be happy to use it as defaults in the task template, but for the moment it does not impact scheduling decisions.

walnut/converter/dump.json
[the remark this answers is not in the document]
ktf: I now see that this part is not exposed, as it was meant to be populated by the Control. Maybe I should add the list of channels for you to consume?

core/task/class.go
```diff
@@ -81,47 +82,104 @@ func (c *Class) UnmarshalYAML(unmarshal func(interface{}) error) (err error) {
...
+	}
+
+	// Flatten bind struct to have channel elements and addressing together
+	for _, bind := range c.Bind {
+		auxBind := _bind{
+			Name:        bind.Name,
+			Type:        bind.Type,
+			SndBufSize:  bind.SndBufSize,
+			RcvBufSize:  bind.RcvBufSize,
+			RateLogging: strconv.Itoa(bind.RateLogging),
```
teo, line 139: These 3 are optional, if the DPL dump doesn't provide explicit values, we can simply omit them from the YAML output. This is not mandatory to fix now, you can make a JIRA ticket so we don't forget about it.

walnut/converter/converter.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// return pointer to string
+func createString(x string) *string {
+	return &x
+}
+
+// ExtractTaskClasses takes in a DPL Dump string and extracts
+// an array of Tasks
+func ExtractTaskClasses(DPL Dump) (tasks []*task.Class, err error) {
```
teo, line 57: Local variable identifiers are normally lowercase... how about `dpl` or `dplDump` instead of `DPL`?

walnut/converter/converter.go
```diff
@@ -0,0 +1,156 @@
...
+
+	for index := range DPL.Workflows {
+		var channelName string
+		if index+1 == len(DPL.Workflows) {
+			channelName = DPL.Workflows[index].Name
+		} else {
+			channelName = "from_" + DPL.Workflows[index].Name + "_to_" + DPL.Workflows[index+1].Name
+		}
```
teo, line 65: I expect this will be changed once Giulio exports channel names in the DPL dump.

walnut/converter/converter.go
```diff
@@ -0,0 +1,156 @@
...
+				Type:        channel.ChannelType("push"), // defaulting to push
+				SndBufSize:  1000,
+				RcvBufSize:  1000,
+				RateLogging: 60,
+				Transport:   channel.TransportType("shmem"),
+			},
+			Addressing: "ipc",
+		}
+
+		// Not required for Task Templates
```
teo, line 80: It's true it's not required, but I'd still suggest including default values for all channels in the TTs, including `connect` (i.e. outbound) ones. The target is the only bit here that cannot be set in the TT.
Not urgent, can also be a JIRA ticket.

walnut/converter/converter.go
```diff
@@ -0,0 +1,156 @@
...
+			Defaults: gera.MakeStringMapWithMap(map[string]string{
+				"user": "flp",
+			}),
+			Control: struct {
+				Mode controlmode.ControlMode "yaml:\"mode\""
+			}{Mode: controlmode.FAIRMQ},
+			Command: &common.CommandInfo{
+				// Env: []string, -> Default to empty array
+				Shell:     createBool(true),
+				Value:     &DPL.Metadatas[index+1].Executable,
```
teo, line 108: "Data" and its compounds are singular uncountable. Please use "Metadata" instead of "Metadatas".

walnut/converter/converter.go
```diff
@@ -0,0 +1,156 @@
...
+	return &x
+}
+
+// ExtractTaskClasses takes in a DPL Dump string and extracts
+// an array of Tasks
+func ExtractTaskClasses(DPL Dump) (tasks []*task.Class, err error) {
+
+	for index := range DPL.Workflows {
+		var channelName string
+		if index+1 == len(DPL.Workflows) {
```
teo, line 61: I've noticed you use `index+1` multiple places, looks like a magic value. What's its significance, just navigating through components of the DPL dump? Perhaps it could be a variable instead of doing `index+1` multiple times.

walnut/converter/converter.go
```diff
@@ -0,0 +1,156 @@
...
+				Ports:  task.Ranges{}, //begin - end OR range
+			},
+			Bind: []channel.Inbound{defaultBindChannel},
+			Properties: gera.MakeStringMapWithMap(map[string]string{
+				"severity": "trace",
+				"color":    "false",
+			}),
+			// Connect:   []channel.Outbound{defaultConnectChannel},
+		}
+		// fmt.Printf("\nTASK:\n%v\n", task)
```
teo, line 124: I propose using `common/logger` and related `logrus` facilities instead of `fmt` for logging output. See https://github.com/AliceO2Group/Control/blob/master/coconut/control/control.go for some examples.

If you ever need to dump a complex struct for debugging purposes, you can also use `k0kubun/pp`. Examples can be found here https://github.com/AliceO2Group/Control/blob/master/core/workflow/load.go and elsewhere in AliECS.

Not urgent & can be a JIRA ticket for you.

walnut/converter/converter.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	for _, SingleTask := range extractedTasks {
+		YAMLData, err := yaml.Marshal(SingleTask)
+		if err != nil {
+			return fmt.Errorf("marshal failed: %w", err)
+		}
+
+		// Write marshaled YAML to file
+		err = ioutil.WriteFile("tasks/"+SingleTask.Identifier.Name+".yaml", YAMLData, 0755)
```
teo, line 150: `0755` is appropriate for a directory but not for a file, which should default to `0644`. The reason for this is that the `x` permission means execute for files, but `cd` for directories.

walnut/converter/dump.json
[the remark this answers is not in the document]
teo: @ktf Yes please, this would be most useful. What we do is essentially building the `--channel-config`, but as a tree for `FairMQOptions` instead of as a command line param.

walnut/converter/converter.go
obviyus: The items in the `workflows` array begin with the first `task`. The items in the metadata array additionally have an `internal-dpl-clock` item preceding all the others. This index value is simply an offset to match the correct `workflow` with it's `metadata`. I'll add a variable.

walnut/converter/converter.go
obviyus: Ah, it was simply to differentiate the array of `metadata` from each item. I'll make the change.

core/task/class.go
obviyus: Considering these are currently not present in a DPL dump, how does defaulting them to empty values and adding a `yaml:"omitempty"` to the Marshaler sound? We can simply replace the default values as and when we get an output from the dump.

teo: Agreed. This kind of information should be filled either by an explicit setting in the DPL dump, or by default values passed by the user.

walnut/converter/converter.go
teo: The single item can be `entry`, `metadataEntry`, `metadataItem`, etc.

walnut/converter/converter.go
teo: `internal-dpl-clock` is a task like any other, it should not receive special treatment. If/when DPL drops it, it will simply not appear in dumps any more.

obviyus: Since `internal-dpl-clock` is only present inside the `metadata` array and not the `workflow` array of the DPL dump, the current implementation does not include it in the import. I'm not sure if this a bug or an intentional design choice.

@ktf since `internal-dpl-clock` is a task, should it not be included in the `workflow` array?

walnut/converter/converter.go
obviyus: Issue created: https://alice.its.cern.ch/jira/browse/OCTRL-246

walnut/converter/dump.json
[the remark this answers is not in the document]
ktf: Done in AliceO2Group/AliceO2#3870. Let me know if it helps.

teo:
Superseded by https://github.com/AliceO2Group/Control/pull/135

walnut/converter/converter.go
teo: @ktf could you please comment on this? It seems like `internal-dpl-clock` is treated differently from other tasks in the DPL dump, and only appears in the `metadata` section. Is this an inconsistency of the DPL dump output, or are we missing some understanding of the DPL dump format?

PR #131, opened 2020-06-09
Force destroy operation

miltalex:
Closes OCTRL-233.


executor/executable/controllabletask.go
```diff
@@ -387,17 +392,35 @@ func (t *ControllableTask) Kill() error {
 			"targetList": cmd.TargetList,
 		}).
 		Debug("state DONE not reached, about to commit transition")
+		
+		// Call cmd.Commit() asynchronous
+		commitDone := make(chan *CommitResponse)
+    	go func() {
```
teo, line 398: There seems to be an indentation issue here, not sure if it's tabs or bad IDE formatting. We should always use spaces.

executor/executable/controllabletask.go
```diff
@@ -387,17 +392,35 @@ func (t *ControllableTask) Kill() error {
...
+			var cr CommitResponse
+			cr.newState, cr.transitionError = cmd.Commit()
+			commitDone <- &cr
+		}()
+		
+		// Set timeout cause OCC is locking up so killing is not possible. The following approach
+		// help us to bypass the OCC transition, so we can kill the tasks on force destroy env. 
+		// Currently this will run for every Kill message receive (force or not).
+		// TODO: Find a better way to distinguish force from plain Kill message.
```
teo, line 407: I actually think that this behavior is correct even if `force == false`. On the core side we distinguish between force or not, so we allow or disallow the termination depending on the global state when `force == false`.

But once we get to this `Kill()`, we already know that the preconditions for terminations were satisfied at the core level (either factually or by virtue of `force == true`), so we know that the expectation is for the controlled process to die.

executor/executable/controllabletask.go
```diff
@@ -387,17 +392,35 @@ func (t *ControllableTask) Kill() error {
...
+		var _cr *CommitResponse
+		select {
+		case _cr = <- commitDone:
+		case <-time.After(45 * time.Second):
```
teo, line 411: Since these transitions are just the `Kill` path, let's set this to 15s. No process should take more than this, because we're not pushing values or reconfiguring, it's just cleanup.

executor/executable/controllabletask.go
```diff
@@ -387,17 +392,35 @@ func (t *ControllableTask) Kill() error {
...
+			var cr CommitResponse
+			cr.newState, cr.transitionError = cmd.Commit()
+			commitDone <- &cr
+		}()
+		
+		// Set timeout cause OCC is locking up so killing is not possible. The following approach
+		// help us to bypass the OCC transition, so we can kill the tasks on force destroy env. 
+		// Currently this will run for every Kill message receive (force or not).
+		// TODO: Find a better way to distinguish force from plain Kill message.
+		var _cr *CommitResponse
```
teo, line 408: Leading underscore is not really a convention in Go, so it confused me a bit to see it here. Its usefulness is mostly if you want to have a type name with a leading capital, but keep it unexported (package private), e.g. `_MyType` instead of `MyType`, or perhaps for function-local type definitions. I propose `commitResponse` instead.

executor/executable/controllabletask.go
```diff
@@ -387,17 +392,35 @@ func (t *ControllableTask) Kill() error {
...
+		select {
+		case _cr = <- commitDone:
+		case <-time.After(45 * time.Second):
+			log.Error("deadline exceeded")
+			break
+    	}
```
teo, line 414: Indentation.

executor/executable/controllabletask.go
miltalex: Okay, I will remove the comment.

PR #135, opened 2020-06-22
WIP: [walnut] load DPL Dump in workflow.Role

obviyus:
Load DPL Dumps generated by `O2Suite/latest` into a `workflow.Role`


core/workflow/loadDPL.go
```diff
@@ -0,0 +1,65 @@
...
+					mu:     sync.RWMutex{},
+					status: 0,
+				},
+				state: SafeState{
+					mu:    sync.RWMutex{},
+					state: 0,
+				},
+				Defaults: gera.MakeStringMapWithMap(taskItem.Defaults.Raw()),
+				Vars:     &gera.StringWrapMap{},
+				UserVars: &gera.StringWrapMap{},
```
obviyus, line 54: @teo: which would be the most appropriate field to store the `Class.Properties` StringMap?

core/workflow/loadDPL.go
```diff
@@ -0,0 +1,65 @@
...
+				Locals:   nil,
+				Bind:     taskItem.Bind,
+			},
```
obviyus, line 57: @teo: let me know which fields we can drop.

core/workflow/loadDPL.go
teo: In `task.Class` it's a `gera.StringMap` (backed by a `map[string]string`). Elsewhere, none.

Please note that `workflow.role*` types and `task.Class` represent different entities. There is no `Properties` field associated with `roleBase` (or any of the `workflow.Role`-related types).

In a `task.Class`, `Properties` means "push these variables as parameters during the `CONFIGURE` transition". The roles tree can define variables, but only the `task.Class` can say "this is the format in which variables `x`, `y` and `z` must be presented to the controlled process".

core/workflow/loadDPL.go
teo: At least the following fields should never appear in a workflow template file:
* `parent`
* `status`
* `state`
* `UserVars`
* `Locals`

Furthermore, I expect `Defaults` and `Vars` to be empty at least initially. If they are empty, they can be omitted.
`Constraints` can be empty for now, but one of the more important tasks of this summer will be to figure out how to generate them.

core/workflow/loadDPL.go
```diff
@@ -0,0 +1,65 @@
...
+
+func LoadDPL(tasks []*task.Class) (workflow Role, err error) {
+	root := new(aggregatorRole)
+
+	for _, taskItem := range tasks {
+		SingleTaskRole := taskRole{
+			roleBase: roleBase{
+				Name:        taskItem.Identifier.Name,
+				parent:      root,
+				Connect:     taskItem.Connect,
```
teo, line 42: Careful, in Go maps and slices are pointer types. You are not making a copy.

I suggest making this, as well as the `Constraints` field, initialized to empty values instead of making them point to the `task.Class`. Nothing should have a pointer to `task.Class` or its fields except the (indirect) `Task` pointer in `taskRole`.

core/workflow/loadDPL.go
```diff
@@ -0,0 +1,65 @@
...
+				Constraints: taskItem.Constraints,
+				status: SafeStatus{
+					mu:     sync.RWMutex{},
+					status: 0,
+				},
+				state: SafeState{
+					mu:    sync.RWMutex{},
+					state: 0,
+				},
+				Defaults: gera.MakeStringMapWithMap(taskItem.Defaults.Raw()),
```
teo, line 52: As before, let's start from an empty map instead of copying from `task.Class`. A simple `gera.MakeStringMap()` should do the trick.

core/workflow/loadDPL.go
```diff
@@ -0,0 +1,65 @@
...
+				Vars:     &gera.StringWrapMap{},
+				UserVars: &gera.StringWrapMap{},
+				Locals:   nil,
+				Bind:     taskItem.Bind,
```
teo, line 56: Here too, let's not point to the `task.Class.Bind`, there is no benefit to doing this, and much potential for confusion.

core/workflow/loadDPL.go
obviyus: I see, thanks for the clarification.

core/workflow/loadDPL.go
```diff
@@ -38,25 +36,20 @@ func LoadDPL(tasks []*task.Class) (workflow Role, err error) {
...
-				Defaults: gera.MakeStringMapWithMap(taskItem.Defaults.Raw()),
-				Vars:     &gera.StringWrapMap{},
-				UserVars: &gera.StringWrapMap{},
-				Locals:   nil,
-				Bind:     taskItem.Bind,
+				Connect:     nil,
+				Constraints: nil,
+				Defaults:    gera.MakeStringMap(),
+				Vars:        gera.MakeStringMap(),
+				Bind:        nil,
```
teo, line 43: Just to clarify: the question that resulted in this change was which fields should not appear in the template file, not which should be left empty. The difference is subtle but could be important.
Specifically `parent` is something you probably want to still set to `root`, even though it doesn't end up in the WFT, because when generating channel paths you might need to be able to navigate the workflow tree (`GetParent()`/`GetPath()` functions and related).

core/workflow/loadDPL.go
```diff
@@ -0,0 +1,58 @@
...
+				Name:        taskItem.Identifier.Name,
+				Connect:     nil,
+				Constraints: nil,
+				Defaults:    gera.MakeStringMap(),
+				Vars:        gera.MakeStringMap(),
+				Bind:        nil,
+			},
+		}
+
+		SingleTaskRole.Connect = taskItem.Connect
```
teo, line 47: As before, please take care not to do this.
`Connect` here is a slice, which means it behaves like a pointer type. You are not copying the `Connect` entries, you are just passing the pointer. This is not what we want. `Connect` entries should be generated, and the `task.Class`'s `Connect` slice can be used as base for this, but if so, it needs to be a deep copy and not a pointer copy.

You can achieve this with `append`.

core/workflow/loadDPL.go
```diff
@@ -0,0 +1,58 @@
...
+		SingleTaskRole.Constraints = taskItem.Constraints
```
teo, line 48: Same as `Connect`.

core/workflow/loadDPL.go
```diff
@@ -0,0 +1,58 @@
...
+		SingleTaskRole.Defaults = gera.MakeStringMapWithMap(taskItem.Defaults.Raw())
+		SingleTaskRole.Bind = taskItem.Bind
```
teo, line 50: Same as `Connect`.

core/workflow/loadDPL.go
obviyus: My bad, I misunderstood. I thought assigning during declaration of the `SingleTaskRole` meant that the pointer would be passed. I expected assigning it outside the declaration to be different. I'll have this fixed.

core/workflow/loadDPL.go
obviyus: I see, of these which ones should we still keep in the `roleBase` declaration (for example `parent` from your recent comment)? Since we're defaulting them to empty, I thought not explicitly assigning them would yield the same result. Omission from the WFT can be done through the `Marshaller` type.

core/workflow/loadDPL.go
```diff
@@ -36,20 +36,32 @@ func LoadDPL(tasks []*task.Class) (workflow Role, err error) {
...
 
-		SingleTaskRole.Connect = taskItem.Connect
-		SingleTaskRole.Constraints = taskItem.Constraints
+		for _, eachConnect := range taskItem.Connect {
+			SingleTaskRole.Connect = append(SingleTaskRole.Connect, eachConnect)
+		}
+
+		for _, eachConstraint := range taskItem.Constraints {
+			SingleTaskRole.Constraints = append(SingleTaskRole.Constraints, eachConstraint)
+		}
```
teo, line 56: Your approach is correct for a manual deep copy, but in Go you can use a simpler single statement with the `...` operator to replace the whole loop:
`SingleTaskRole.Connect = append(SingleTaskRole.Connect, taskItem.Connect...)`

This improves readability because one can immediately tell it's a straight deep copy without in-loop manipulation of each entry.

core/workflow/loadDPL.go
```diff
@@ -36,20 +36,32 @@ func LoadDPL(tasks []*task.Class) (workflow Role, err error) {
...
+
 		SingleTaskRole.Defaults = gera.MakeStringMapWithMap(taskItem.Defaults.Raw())
-		SingleTaskRole.Bind = taskItem.Bind
-		SingleTaskRole.Task = task.ClassToTask(taskItem, &SingleTaskRole)
 
+		for _, eachBind := range taskItem.Bind {
+			SingleTaskRole.Bind = append(SingleTaskRole.Bind, eachBind)
+		}
```
teo, line 62: See previous comment.

core/workflow/loadDPL.go
obviyus: Perfect! That looks much better

PR #138, opened 2020-06-27
[walnut] Marshal workflow.Role to YAML

obviyus:
Closes OCTRL-256.


core/workflow/loadDPL.go
```diff
@@ -57,5 +62,95 @@ func LoadDPL(tasks []*task.Class) (workflow Role, err error) {
 	}
 
 	workflow = root
```
obviyus, line 64: How should we decide a `name` for the `root` role? The DPL is only information about the constituent tasks. Possible choices:
- use the name of the executable
- executable name along with some logic to derive a `workflow` name based on the constituent tasks

teo: I don't think inferring a name from the executables would be reliable or useful.

What about using the filename of the input DPL dump?

obviyus: Is there a pattern for the input DPL Dumps? From my experience with O2, it’s the user who provides the name for the DPL Dump.

I’m not against this idea. But this could potentially end up with workflow names being irrelevant to the content.

walnut/converter/converter.go
```diff
@@ -139,19 +144,50 @@ func TaskToYAML(extractedTasks []*task.Class) (err error) {
...
 
 	for _, SingleTask := range extractedTasks {
-		YAMLData, err := yaml.Marshal(SingleTask)
+		YAMLData, err := yaml.Marshal(&SingleTask)
 		if err != nil {
 			return fmt.Errorf("marshal failed: %v", err)
 		}
 		fileName := "tasks/" + SingleTask.Identifier.Name + ".yaml"
 		f, err := os.Create(fileName)
-		defer f.Close()
```
teo, line 148: This was actually better than moving the `f.Close()` to the bottom, because now if you hit the `err` case you're leaving an open file handle.

walnut/converter/converter.go
```diff
@@ -139,19 +144,50 @@ func TaskToYAML(extractedTasks []*task.Class) (err error) {
...
+		}
+	}
+
+	yamlDATA, err := workflow.RoleToYAML(input)
+	if err != nil {
+		return fmt.Errorf("error converting role to YAML: %v", err)
+	}
+
+	var fileName = "workflows/"
+	if input.GetName() == "" {
```
teo, line 181: I think we should error out if there's no input name. How could this ever happen?

walnut/converter/converter.go
```diff
@@ -139,19 +144,50 @@ func TaskToYAML(extractedTasks []*task.Class) (err error) {
...
+		fileName += "no-name.yaml"
+	} else {
+		fileName += input.GetName()
```
teo, line 184: Does this take into account that the input is named `*.json` and the output `*.yaml`?

core/task/class.go
```diff
@@ -98,35 +98,14 @@ func (c *Class) UnmarshalYAML(unmarshal func(interface{}) error) (err error) {
...
-			Mode string                     `yaml:"mode"`
+		Name        string                  `yaml:"name"`
+		Defaults    map[string]string       `yaml:"defaults"`
+		Control     struct {
+			Mode    string                  `yaml:"mode"`
 		}                                   `yaml:"control"`
 		Wants       ResourceWants           `yaml:"wants"`
-		Bind        []_bind                 `yaml:"bind"`
-		Connect 	[]_connect              `yaml:"connect"`
+		Bind        []*channel.Inbound      `yaml:"bind"`
```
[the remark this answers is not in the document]
obviyus: I recently switched all `MarshalYAML` receivers to pointers (mimicking how `UnmarshalYAML` works in rest of `Control` ). To that effect, I appropriately converted all the types being marshaled to their pointer equivalents as well.

core/workflow/loadDPL.go
```diff
@@ -57,5 +62,95 @@ func LoadDPL(tasks []*task.Class) (workflow Role, err error) {
...
 	return workflow, nil
 }
+
+// Aux struct to fulfil export requirement by yaml.Marshal
+type auxAggregatorRole struct {
+	RoleBase   roleBase
+	Aggregator aggregator
+}
+
+func RoleToYAML(input Role) ([]byte, error) {
```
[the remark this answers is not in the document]
obviyus: Trying to use `Role` as a receiver for `MarshalYAML` yields an `Invalid receiver type 'Role'`. I decoupled the `RoleToYAML` function to pass `auxRole` as the receiver for Marshller. It does seem a bit hacky. Let me know if there's a better way to approach this.

walnut/converter/converter.go
obviyus: I expect to strip `*.json` from the name at the time of conversion from DPL Dump to Role.

core/task/class.go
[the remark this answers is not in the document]
teo: I might be wrong about this but I think you don't need to do this for slice types, i.e. I reckon that `[]channel.Inbound` should anyway trigger `(* Inbound) UnmarshalYAML`.

core/workflow/loadDPL.go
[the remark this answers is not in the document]
teo: Of course, note that I never suggested `(Role) MarshalYAML` as I'm aware that this is impossible. But I assert that `(* aggregatorRole) MarshalYAML` is possible, and should be enough without additional, non-local auxiliary types (also related to the discussion on using `map[string]interface{}` as auxiliary format.

core/task/class.go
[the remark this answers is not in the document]
obviyus: Unfortunately, marshalling a `[]channel.Inbound` does not trigger the `(i Inbound) MarshalYAML`. Changing it to `(i *Inbound) MarshalYAML` works fine. I'll go ahead and make the change for `channel` types.

walnut/converter/converter.go
obviyus: Fair point. I thought having an `defer` inside a for loop is [a bad idea](https://blog.golang.org/defer-panic-and-recover), but the orphaned file handle is probably worse. I'll have it fixed.

core/workflow/aggregatorrole.go
```diff
@@ -64,6 +65,42 @@ func (r *aggregatorRole) UnmarshalYAML(unmarshal func(interface{}) error) (err e
 	return
 }
 
+func (r *aggregatorRole) MarshalYAML() (interface{}, error) {
+	// Aux struct to cast Roles as an array
+	type rootWorkflow struct {
```
teo, line 70: If it's an auxiliary struct I'd rather it be called something other than `rootWorkflow`. The reason is that "the root role of the workflow tree" already has a specific meaning (see `load.go`).

Generally, a role that's at the root of a workflow is necessarily of type `aggregatorRole`, but `aggregatorRole`s can also occur elsewhere in a workflow. See https://github.com/AliceO2Group/ControlWorkflows/blob/master/workflows/readout-stfb-qc.yaml#L28 for an example of this.

Post-MVP, an effort to implement workflow composition i.e. attaching a converted DPL workflow onto an existing AliECS workflow essentially boils down to appending subworkflow structures onto the `roles` field of an existing `aggregatorRole`.

core/task/class.go
```diff
@@ -140,31 +119,8 @@ func (c *Class) MarshalYAML() (interface{}, error) {
...
-			SndBufSize:  connect.SndBufSize,
-			RcvBufSize:  connect.RcvBufSize,
-			RateLogging: strconv.Itoa(connect.RateLogging),
-			Transport:   connect.Transport,
-			Target:      connect.Target,
-		}
-		aux.Connect = append(aux.Connect, auxConnect)
+	for _, eachBind := range c.Bind {
+		aux.Bind = append(aux.Bind, &eachBind)
 	}
```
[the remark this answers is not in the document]
obviyus: I'm not sure about the advantages of having an `(i *Inbound)` vs. `(i Inbound)`. Either can be implemented easily.

core/workflow/aggregatorrole.go
```diff
@@ -64,6 +65,42 @@ func (r *aggregatorRole) UnmarshalYAML(unmarshal func(interface{}) error) (err e
...
+		Name     string                   `yaml:"name"`
+		Defaults gera.StringMap           `yaml:"defaults"`
+		Roles    []map[string]interface{} `yaml:"roles"`
+	}
+
+	// TODO: Handle multiple Roles
+	var finalAux []map[string]interface{}
```
teo, line 77: I suspect you don't need this `finalAux` trick. Your output of `aggregator.MarshalYAML` is an `interface{}` with a `map[string]interface{}` inside. You correctly type-assert this and call it `mapAggregator`. This should be enough, because you'll have a map whose only key is `roles`, with a value of type `interface{}` and true type `[]Role`.

You shouldn't have to cast all items to `[]Role` or `[]map[string]interface{}`, because the kv-copy on L90 should be sufficient, and you should then be able to simply output `aux`.

core/workflow/loadDPL.go
```diff
@@ -25,13 +25,21 @@
...
 
-func LoadDPL(tasks []*task.Class) (workflow Role, err error) {
+func LoadDPL(tasks []*task.Class, filename string) (workflow Role, err error) {
+	// FIXME: base roleBase of root defaults to all empty values
 	root := new(aggregatorRole)
 
+	if strings.Contains(filename, ".json"){
+		root.roleBase.Name = filename[:len(filename)-5]
+	}
+
```
teo, line 42: I think generating the filename and/or root role name should happen in the caller, outside the `workflow` package. So `func LoadDPL(tasks []*task.Class, rootRoleName string) (workflow Role, err error)`.

core/workflow/loadDPL.go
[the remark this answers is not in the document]
teo: I find these two `RoleToYAML` functions a bit confusing. This one accepts a `Role` as input, which internally is presumably an `*aggregatorRole`. Instead of creating a new `aggregatorRole` here only to call `yaml.Marshal`, you can type-assert your `input` to `*aggregatorRole` instead.

walnut/converter/converter.go
```diff
@@ -56,38 +58,11 @@ func createString(x string) *string {
 
 // ExtractTaskClasses takes in a DPL Dump string and extracts
 // an array of Tasks
-func ExtractTaskClasses(dplDump Dump) (tasks []*task.Class, err error) {
-
+func ExtractTaskClasses(dplDump Dump, module []string) (tasks []*task.Class, err error) {
```
teo, line 61: Just a minor nitpick, since it's a slice I'd call it `envModules`.

walnut/converter/converter.go
```diff
@@ -111,13 +85,58 @@ func ExtractTaskClasses(dplDump Dump) (tasks []*task.Class, err error) {
...
 			},
-			Bind: bind,
 			Properties: gera.MakeStringMapWithMap(map[string]string{
 				"severity": "trace",
 				"color":    "false",
 			}),
-			Connect: connect,
 		}
+
+		module = append(module, "Control-OCCPlugin")
```
teo, line 94: I'd prefer if `Control-OCCPlugin` was always the last module in the slice.

walnut/converter/converter.go
```diff
@@ -111,13 +85,58 @@ func ExtractTaskClasses(dplDump Dump) (tasks []*task.Class, err error) {
...
+		loadModule := "eval `aliswmod load "
```
teo, line 95: I advise using `fmt.Sprintf` and `strings.Join` for more idiomatic string building.

walnut/converter/converter.go
```diff
@@ -111,13 +85,58 @@ func ExtractTaskClasses(dplDump Dump) (tasks []*task.Class, err error) {
...
+				// String manipulation to generate channel target of the form:
+				// {{ Parent().Path }}.taskName:from_{initiator}_to_{receiver}
+				initiator := channelName[5:strings.Index(channelName, "_to")]
+				singleConnect := channel.Outbound{
+					Channel: channel.Channel{
+						Name:      channelName,
+						Type:      channel.ChannelType("pull"),
+						Transport: channel.TransportType("shmem"),
+					},
+					Target: "{{ Parent().Path }}." + initiator + ":" + channelName,
```
teo, line 134: Advising `fmt.Sprintf` as before, though here it's less critical.

walnut/converter/converter.go
```diff
@@ -155,3 +174,36 @@ func TaskToYAML(extractedTasks []*task.Class) (err error) {
 	}
 	return
 }
+
+func RoleToYAML(input workflow.Role) (err error) {
```
teo, line 178: In light of the fact that you're doing filesystem operations specific to the WFT/TT configuration structure, and to avoid confusion with the other `RoleToYAML` function, I think this one could be called `GenerateWorkflowTemplate`. Similar considerations apply to `TaskToYAML`.

Generally I'd like to have the word `YAML` in functions that deal specifically with `YAML` representation, whereas higher level wrapper functions that handle filesystem and I/O aspects (as opposed to the details of YAML syntax) should have other names. It is known by design that the syntax of these files happens to be YAML, and there's no gain in making it explicit throughout the code path as long as we know what's a TT and what's a WFT. Compare this with `UnmarshalYAML` vs. `workflow.Load`, where the latter deals with a higher level of I/O and template processing, and internally calls `yaml.Unmarshal`.

Also please ensure that you only export the functions you need to export (I haven't checked if this one should be changed).

walnut/converter/converter.go
obviyus: Is that not what's happening here? `module` is a `[]string` provided to this function. Whether it's empty or not, `Control-OCCPlugin` is appended to it.

core/task/channel/outbound.go
```diff
@@ -58,6 +58,35 @@ func (outbound *Outbound) UnmarshalYAML(unmarshal func(interface{}) error) (err
...
 }
 
+func (outbound *Outbound) MarshalYAML() (interface{}, error) {
+	// Flatten Outbound to have Target and Channel elements on the same "level"
+	type _outbound struct {
+		Name        string        `yaml:"name"`
+		Type        ChannelType   `yaml:"type"`
+		SndBufSize  int           `yaml:"sndBufSize,omitempty"`
+		RcvBufSize  int           `yaml:"rcvBufSize,omitempty"`
+		RateLogging string        `yaml:"rateLogging,omitempty"`
```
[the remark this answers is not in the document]
obviyus: I ended up going with `int` for `RateLogging` to maintain consistency with `SndBufSize` and `RcvBufSize`.

core/workflow/aggregatorrole.go
obviyus: Changed `rootWorkflow` to `auxAggregatorRole` (since this comprises `roleBase` values and an `aggregator`).

core/workflow/aggregatorrole.go
obviyus: Indeed, that was my first implementation as well. But it poses a couple of problems:
- if I use the type asserted `map[string]interface{}` as the value for `output.Roles` I get a YAML file with: 
```yaml
roles:
  name:
  connect:
  roles:
    - name:
      connect:
```
Trying to use this to create an env in `coconut` causes an error saying: `cannot unmarshal !!map to workflow._roleUnion`. After having a look at `_roleUnion`, it seems to me that the first `roles` field needs to be an array. Which means we need:
```yaml
roles:
  - name:
    connect:
    roles:
      - name:
        connect:
```
Even if the first `roles` field has a single element. If I try to marshal a `map[string]interface{}` it results in the first output. Using this trick with a `[]map[string]interface{}` results in the second. It is also why an auxiliary `_aggregatorRole` struct is needed. 

This was a hack to make the final YAML mimic the desired one more closely. Let me know if there's a better approach.

walnut/converter/converter.go
obviyus: Excellent suggestion. This converted my 10 line implementation into a single line while also getting rid of the `if` statement.

walnut/converter/converter.go
obviyus: Understood. I'll make the changes.

core/task/class.go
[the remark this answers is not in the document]
obviyus: I decided go drop pointer receivers for all `channel` types.

core/workflow/loadDPL.go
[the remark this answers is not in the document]
obviyus: This is fantastic! Now the `RoleToYAML` function exists solely as a wrapper for `yaml.Marshal` which allows us to avoid exporting `aggregatorRole`.

walnut/converter/converter.go
teo: You're right, my bad.

core/workflow/aggregatorrole.go
teo: Note that I did not suggest to use the type-asserted `mapAggregator := auxAggregator.(map[string]interface{})` as the value of `output.Roles`, I suggested to use `mapAggregator["roles"]`, of type `interface{}` as the value of `output["roles"]` (where `output` is a `map[string]interface{}` too). Did this approach result in trouble?

core/task/channel/inbound.go
```diff
@@ -60,34 +60,29 @@ func (inbound *Inbound) UnmarshalYAML(unmarshal func(interface{}) error) (err er
...
-	}
+auxInbound := _inbound{
+	Name:        inbound.Channel.Name,
+	Type:        inbound.Channel.Type,
+	SndBufSize:  inbound.Channel.SndBufSize,
+	RcvBufSize:  inbound.Channel.RcvBufSize,
+	RateLogging: inbound.RateLogging,
+	Transport:   inbound.Channel.Transport,
+	Addressing:  inbound.Addressing,
+}
```
teo: Indentation issue? It's fine in `_outbound` but not here.

core/task/class.go
```diff
@@ -101,28 +101,26 @@ func (c *Class) MarshalYAML() (interface{}, error) {
 	type _class struct {
 		Name        string                  `yaml:"name"`
 		Defaults    map[string]string       `yaml:"defaults"`
```
teo, line 103: Some of these can be tagged `omitempty`, including `Defaults`, `Bind`, and `Properties`.

obviyus: I did try that but it seems to have no effect. A `default: []` keeps getting marshalled regardless. Similar issues with the others. I'll go ahead an tag them though.

obviyus: For example, this is the output when tagged with `omitempty`:
```yaml
connect: []
constraints: []
defaults: null
```

I'll make a JIRA issue.

PR #139, opened 2020-07-07
[walnut] improved CLI functionality + documentation

obviyus:
Fixes OCTRL-259.


walnut/README.md
```diff
@@ -1,4 +1,76 @@
...
+A successful validation will return no output while any errors will cause the program to exit.
+
+#### SEE ALSO
+
+* [walnut check](./doc/walnut_check.md)
+
+
+### Validating a file
+
+The goal of `walnut` is to convert O² generated dumps to the required number of task templates and workflow templates. The
```
teo, line 59: Since AliECS is part of O², I think this should be reworded to "DPL generated dumps" or "O²/DPL generated dumps".

walnut/README.md
```diff
@@ -1,4 +1,76 @@
...
+Example usage:
+```bash
+$ walnut convert --filename dump.json --modules TestValue1 TestValue2 
+```
+
+A successful validation will return no output while any errors will cause the program to exit.
+
+#### SEE ALSO
+
+* [walnut check](./doc/walnut_convert.md)
```
teo, line 76: `walnut check` here points to `walnut_convert.md` :eyes: 

Generally the `README` is very good, just please do another proofreading pass.

walnut/cmd/check.go
```diff
@@ -37,25 +38,18 @@ import (
 var checkCmd = &cobra.Command{
 	Use:   "check",
 	Short: "check the file passed against a specified schema.",
```
teo, line 40: `validate the file against a specified schema`

Also please make sure that all your `Short` descriptions don't end with a period. IIRC the Cobra docs have some advice on wording (though now I can't find it).

walnut/cmd/check.go
```diff
@@ -37,25 +38,18 @@ import (
...
-schema. This file can be a task template, a workflow template or an O² 
-DPL Dump. Each of those have a schema provided to validate against. 
-
-Usage:
-  walnut check --format [format] [file]
-
-Example:
-  walnut check --format workflow readout-sftb.yaml
+	Long: `The check command validates the given file against a specified schema. This file can be a task template, 
+a workflow template or an O² DPL Dump. Each of those have a schema provided to validate against.
```
teo, line 42: `s/those/these/`
`s/provided //`

walnut/cmd/check.go
```diff
@@ -37,25 +38,18 @@ import (
...
 
 Valid schemata formats:
```
teo, line 44: `s/schemata/schema/` for the same reason you'd say "valid workflow formats" and not "valid workflows formats".

walnut/cmd/check.go
```diff
@@ -37,25 +38,18 @@ import (
...
   workflow  task  dpl_dump`,
 
 	Run: func(cmd *cobra.Command, args []string) {
-		// fmt.Println("check called with arg: " + strings.Join(args, " "))
-
 		format, _ := cmd.Flags().GetString("format")
-		for _, filename := range args {
-			file, err := ioutil.ReadFile(filename)
+		filename, _ := cmd.Flags().GetStringArray("filename")
```
teo, line 49: Why is `filename` a `StringArray`? Do you validate multiple DPL dumps per call? I don't remember if we discussed this.

walnut/cmd/convert.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"io/ioutil"
+	"os"
+)
+
+// convertCmd represents the convert command
+var convertCmd = &cobra.Command{
+	Use:   "convert",
+	Short: "Converts a DPL Dump to the required formats.",
```
teo, line 40: Should not start with a capital letter nor end with a period.

walnut/cmd/convert.go
```diff
@@ -0,0 +1,87 @@
...
+	Long: `The convert command takes a DPL input and outputs task and workflow templates. Optional flags can be provided to
+specify which modules were used during dump generation. By default, both templates are produced. Control-OCCPlugin is
```
teo, line 42: `s/were used during dump generation/should be used when generating task templates/`

What do you mean with "By default, both templates are produced"? Is there an option to only produce TTs or WFTs?

walnut/cmd/convert.go
```diff
@@ -0,0 +1,87 @@
...
+		modules, _ := cmd.Flags().GetStringArray("modules")
+
+		for _, dump := range filename {
+			file, err := ioutil.ReadFile(dump)
+			if err != nil {
+				fmt.Errorf("failed to open file &s: &v", file, err)
+				os.Exit(1)
+			}
+
+			dplDump, err := converter.JSONImporter(file)
```
teo, line 57: Following up on the observations concerning the word `YAML` in function names, I think this one should be a `DPLImporter`. Yes, a DPL dump happens to use JSON as syntax, but this is of secondary importance to the fact that it's a **DPL** dump (i.e. a file with DPL semantics) that we're importing.

walnut/cmd/check.go
obviyus: Yep! I thought it translated naturally into converting a batch of DPL Dumps into their equivalent templates. An example use case: `walnut convert -f dump1.json dump2.json dump3.json`. Right now there is no accommodation for separate folders per dump. This could go either:
- unified `tasks` and `workflows` folders for all input dumps
- separate folders where the parent is the name of the dump

And this works just as well for a single DPL Dump.

walnut/cmd/convert.go
obviyus: My bad, I had a `--template` flag in a previous version which would allow the user to selectively generate either `task` or `workflow` templates. I removed that temporarily because I had some trouble with the flag value check. We can either make a JIRA issue for this to come back at a later date or re-implement it now. Either works for me.

walnut/cmd/check.go
teo: In order to run a workflow in AliECS, one has to anyway commit the WFT and TTs to a Git repo, so in any case I'd prefer 1 target dir (with the usual `tasks` and `workflows` subdirectories) even with multiple DPL inputs.

walnut/cmd/convert.go
teo: I don't think there's much of a use case for outputting only TTs or only WFTs. The output of a DPL dump should always be 1 WFT + an as-needed number of TTs.

walnut/cmd/convert.go
obviyus: Fixed in #138

walnut/README.md
```diff
@@ -1,4 +1,76 @@
...
 The O² **w**orkflow **a**dministration and **l**i**n**ting **ut**ility is a command line program for workflow configuration
-and import validation.
\ No newline at end of file
+and import validation.
+
+## Configuration file
+
+You can check the local `walnut` configuration with
+```bash
+$ walnut about
```
teo, line 10: This command is not implemented, is it? Ok to keep the mention if you plan on implementing it soon after merging.

walnut/README.md
```diff
@@ -1,4 +1,76 @@
...
+
+Available Commands:
+  check       check the file passed against a specified schema.
+  convert     Converts a DPL Dump to the required formats.
+  help        Help about any command
+
+Flags:
+      --config string   optional configuration file for walnut (default $HOME/.config/walnut/settings.yaml)
+  -h, --help            help for walnut
+  -t, --toggle          Help message for toggle
```
teo, line 34: What's toggle?

walnut/README.md
```diff
@@ -1,4 +1,76 @@
...
+
+### Validating a file
+
+To check if a file conforms to the pre-defined schemas, run the `check` command: 
+```bash
+$ walnut check
+```
+Example usage:
+```bash
+$ walnut check --filename dump.json --format workflow
```
teo, line 47: `filename` should be an arg and not a flag, so the example would become
```bash
$ walnut check dump.json --format workflow
```

walnut/README.md
```diff
@@ -1,4 +1,76 @@
...
+command used is `convert`:
+```bash
+$ walnut convert
+```
+
+Convert can also receive optional arguments to specify which modules to consider while generating task templates.
+
+Example usage:
+```bash
+$ walnut convert --filename dump.json --modules TestValue1 TestValue2
```
teo, line 69: Should become
```bash
$ walnut convert dump.json --modules TestValue1 TestValue2
```

walnut/cmd/check.go
```diff
@@ -36,26 +37,19 @@ import (
...
+Valid schema formats:
   workflow  task  dpl_dump`,
 
 	Run: func(cmd *cobra.Command, args []string) {
-		// fmt.Println("check called with arg: " + strings.Join(args, " "))
-
 		format, _ := cmd.Flags().GetString("format")
-		for _, filename := range args {
-			file, err := ioutil.ReadFile(filename)
+		filename, _ := cmd.Flags().GetStringArray("filename")
```
teo, line 49: The filenames should come from cobra args instead of a `--filename` flag.
The best practice for command line programs is `command [args] [flags]` or `command [flags] [args]`:
* arguments express *on what* we want to run the operation
* flags modify *how* we want the operation to be executed.

Of course this doesn't apply cleanly to every problem, but I think in this case it does and should.

walnut/converter/converter.go
```diff
@@ -102,48 +80,89 @@ func ExtractTaskClasses(dplDump Dump) (tasks []*task.Class, err error) {
...
-		// FIXME: use absolute paths
-		errDir := os.Mkdir("tasks", 0755)
-		if errDir != nil {
-			return fmt.Errorf("create dir failed: %v", err)
+func sanitizeCmdLineArgs (input []string, taskName string) (output []string) {
+	for _, value := range input {
+		// Check args for dump arguments and remove them
+		if  strings.Contains(value, "--dump-workflow") ||
+			strings.Contains(value, "--dump-workflow-file") ||
+			strings.Contains(value, ".json") {} else {
```
teo, line 143: Might not be safe to remove all params that end with `.json`.

walnut/README.md
obviyus: Not sure, must've been something added automatically by `cobra`. Removed.

walnut/converter/converter.go
obviyus: I'll remove the check for `.json` but in that case we will have the name of the dump (to which it was exported to) passed onto `cmdLineArgs`.

walnut/README.md
obviyus: Added in 2b52f2a

walnut/converter/converter.go
teo: Not necessarily. You can only remove the first item after `--dump-workflow-file` if this item is `.json`, since you know the input slice is ordered.

walnut/cmd/check.go
```diff
@@ -67,12 +61,15 @@ Valid schemata formats:
...
+	Args: cobra.MinimumNArgs(1),
 }
 
+var format string
+
 func init() {
 	rootCmd.AddCommand(checkCmd)
 
-	checkCmd.Flags().StringP("format", "f", "", "format to validate against")
+	checkCmd.Flags().StringP("format", "", "", "format to validate against")
```
teo, line 72: Why remove the `f` shorthand? It would be nice to have.

walnut/cmd/convert.go
```diff
@@ -0,0 +1,86 @@
...
+
+	Run: func(cmd *cobra.Command, args []string) {
+		for _, dumpFile := range args {
+			// Strip .json from end of filename
+			nameOfDump := dumpFile[:len(dumpFile)-5]
+
+			file, err := ioutil.ReadFile(dumpFile)
+			if err != nil {
+				err = fmt.Errorf("failed to open file &s: &v", dumpFile, err)
+				fmt.Println(err.Error())
```
teo, line 53: Not necessary for this PR, but please make a JIRA ticket to use `common/logger` for such error messages throughout `walnut`. In this case, it would be a `log.Fatal`.

walnut/cmd/convert.go
```diff
@@ -0,0 +1,86 @@
...
+specify which modules should be used when generating task templates. Control-OCCPlugin is always used as module.`,
+
+	Run: func(cmd *cobra.Command, args []string) {
+		for _, dumpFile := range args {
+			// Strip .json from end of filename
+			nameOfDump := dumpFile[:len(dumpFile)-5]
+
+			file, err := ioutil.ReadFile(dumpFile)
+			if err != nil {
+				err = fmt.Errorf("failed to open file &s: &v", dumpFile, err)
```
teo, line 52: Here you are wrapping an existing error coming from `ioutil`, so you should use `%w` instead of `%v`. This applies to several other instances in this file (and maybe elsewhere too).

See https://blog.golang.org/go1.13-errors

walnut/converter/converter.go
```diff
@@ -127,32 +130,38 @@ func ExtractTaskClasses(dplDump Dump, envModules []string) (tasks []*task.Class,
...
-		// FIXME: use absolute paths
-		errDir := os.Mkdir("tasks", 0755)
-		if errDir != nil {
-			return fmt.Errorf("create dir failed: %v", err)
+func sanitizeCmdLineArgs (input []string, taskName string) (output []string) {
+	for _, value := range input {
+		// Check args for dump arguments and remove them
+		if  strings.Contains(value, "--dump-workflow") ||
+			strings.Contains(value, "--dump-workflow-file") {} else {
+				output = append(output, value)
```
teo, line 143: Also remove the entry immediately after `--dump-workflow-file` if said entry is `*.json` as previously mentioned.

walnut/cmd/convert.go
obviyus: Right here: https://alice.its.cern.ch/jira/browse/OCTRL-246. I'll get to this soon

walnut/cmd/convert.go
obviyus: JIRA issue here: https://alice.its.cern.ch/jira/browse/OCTRL-272

PR #140, opened 2020-07-13
[walnut] Allow selection of output directory

obviyus:
Fixes OCTRL-278.

### DIFF
--- a/walnut/cmd/convert.go
+++ b/walnut/cmd/convert.go
@@ -35,6 +35,9 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var outputDir string
+var modules []string
+
 // convertCmd represents the convert command
 var convertCmd = &cobra.Command{
 	Use:   "convert",
@@ -76,12 +79,12 @@ specify which modules should be used when generating task templates. Control-OCC
 	Args: cobra.MinimumNArgs(1),
 }
 
-var modules []string
-
 func init() {
 	convertCmd.Flags().StringArrayVarP(&modules, "modules", "m", []string{}, "modules to load")
 	_ = viper.BindPFlag("modules", convertCmd.Flags().Lookup("modules"))
 
+	convertCmd.PersistentFlags().StringVarP(&outputDir, "output-dir", "o", "",
+		"optional output directory")
 	viper.BindPFlag("output-dir", rootCmd.Flags().Lookup("output-dir"))
 
 	rootCmd.AddCommand(convertCmd)
--- a/walnut/cmd/convert.go
+++ b/walnut/cmd/convert.go
@@ -57,7 +57,7 @@ specify which modules should be used when generating task templates. Control-OCC
 			dplDump, err := converter.DPLImporter(file)
 			taskClass, err := converter.ExtractTaskClasses(dplDump, modules)
 
-			err = converter.GenerateTaskTemplate(taskClass, nameOfDump)
+			err = converter.GenerateTaskTemplate(taskClass, outputDir)
 			if err != nil {
 				err = fmt.Errorf("conversion to task failed for %s: %w", dumpFile, err)
 				fmt.Println(err.Error())
@@ -65,7 +65,7 @@ specify which modules should be used when generating task templates. Control-OCC
 			}
 
 			role, err := workflow.LoadDPL(taskClass, nameOfDump)
-			err = converter.GenerateWorkflowTemplate(role, nameOfDump)
+			err = converter.GenerateWorkflowTemplate(role, outputDir)
 			if err != nil {
 				err = fmt.Errorf("conversion to workflow failed for %s: %w", dumpFile, err)
 				fmt.Println(err.Error())
@@ -81,6 +81,8 @@ var modules []string
 func init() {
 	convertCmd.Flags().StringArrayVarP(&modules, "modules", "m", []string{}, "modules to load")
 	_ = viper.BindPFlag("modules", convertCmd.Flags().Lookup("modules"))
+
+	viper.BindPFlag("output-dir", rootCmd.Flags().Lookup("output-dir"))
 
 	rootCmd.AddCommand(convertCmd)
 }
--- a/walnut/cmd/convert.go
+++ b/walnut/cmd/convert.go
@@ -107,6 +107,7 @@ specify which modules should be used when generating task templates. Control-OCC
 				result := true
 				prompt := &survey.Confirm{
 					Message: "Would you like to view the git diff?",
+					Default: true,
 				}
 				_ = survey.AskOne(prompt, &result)
 
--- a/walnut/cmd/convert.go
+++ b/walnut/cmd/convert.go
@@ -25,7 +25,6 @@
 package cmd
 
 import (
-	"bufio"
 	"fmt"
 	"io/ioutil"
 	"log"
@@ -34,6 +33,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/AlecAivazis/survey/v2"
 	"github.com/mitchellh/go-homedir"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -102,15 +102,17 @@ specify which modules should be used when generating task templates. Control-OCC
 			}
 
 			if isGitRepo {
-				fmt.Print("Press 'Enter' to view git diff...")
-				bufio.NewReader(os.Stdin).ReadBytes('\n')
-				fmt.Print(runGitCmd([]string{"diff"}))
-
-				fmt.Print("Press 'Enter' to view git status...")
-				bufio.NewReader(os.Stdin).ReadBytes('\n')
 				fmt.Printf(runGitCmd([]string{"status"}))
 
-				fmt.Print("You can commit/reset the changes")
+				result := true
+				prompt := &survey.Confirm{
+					Message: "Would you like to view the git diff?",
+				}
+				_ = survey.AskOne(prompt, &result)
+
+				if result {
+					fmt.Printf(runGitCmd([]string{"diff"}))
+				}
 			}
 		}
 	},
--- a/walnut/cmd/convert.go
+++ b/walnut/cmd/convert.go
@@ -25,14 +25,21 @@
 package cmd
 
 import (
+	"bufio"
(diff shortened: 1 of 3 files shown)


walnut/converter/converter.go
```diff
@@ -175,21 +189,57 @@ func GenerateTaskTemplate(extractedTasks []*task.Class, dumpFileName string) (er
...
+		response = strings.ToLower(strings.TrimSpace(response))
+
+		if response == "y" || response == "yes" {
+			return true
+		} else if response == "n" || response == "no" {
+			return false
+		} else {
+			fmt.Printf("Invalid response\n")
+		}
+	}
```
teo, line 212: Your logic is correct but I think we can & should do better than this manual implementation.

Please take a look at https://github.com/AlecAivazis/survey which popped up after a quick search. It handles automatically yes-no questions, as well as many other kinds of interactive terminal input.

You can also look for an alternative package in https://awesome-go.com/ but don't spend too much time on this if `AlecAivazis/survey` turns out to work well.

walnut/converter/converter.go
```diff
@@ -175,21 +189,57 @@ func GenerateTaskTemplate(extractedTasks []*task.Class, dumpFileName string) (er
...
 	os.MkdirAll(path, os.ModePerm)
 
 	yamlDATA, err := workflow.RoleToYAML(input)
 	if err != nil {
 		return fmt.Errorf("error converting role to YAML: %v", err)
 	}
 
-	fileName := filepath.Join(path, input.GetName() + ".yaml")
+	fileName := filepath.Join(path, input.GetName()+".yaml")
+	if _, err := os.Stat(fileName); os.IsNotExist(err) {} else {
```
teo, line 232: Constructs such as `if <condition> { //noop } else { //something }` are difficult to read and should be avoided in favor of `if !<condition> { //something }`.

walnut/converter/converter.go
```diff
@@ -175,21 +189,57 @@ func GenerateTaskTemplate(extractedTasks []*task.Class, dumpFileName string) (er
...
+		if confirmOverwrite(fmt.Sprintf("%s already exists, overwrite?", fileName)) {} else {
+			// return here get shadowed
+			goto EXIT
```
teo, line 235: The situations in Go where the benefits of using `goto` outweigh the drawbacks are rare, and I don't think this is one of them. If your `EXIT` label points to a single `return` statement, why not just use `return`?

obviyus: Ah the `err` value gets shadowed while returning inside the `if` block. Changed it to return a `nil`.

walnut/converter/converter.go
obviyus: I chose to go with: github.com/manifoldco/promptui

walnut/cmd/convert.go
```diff
@@ -57,30 +67,63 @@ specify which modules should be used when generating task templates. Control-OCC
...
+				fmt.Print("Press 'Enter' to view git diff...")
+				bufio.NewReader(os.Stdin).ReadBytes('\n')
+				fmt.Print(runGitCmd([]string{"diff"}))
+
+				fmt.Print("Press 'Enter' to view git status...")
+				bufio.NewReader(os.Stdin).ReadBytes('\n')
+				fmt.Printf(runGitCmd([]string{"status"}))
+
+				fmt.Print("You can commit/reset the changes")
+			}
```
obviyus, line 114: @teo: do you think it's worth doing a clear screen after each?

teo: I don't think so. `walnut` should be an interactive and noninteractive command line utility, not a console GUI application.

walnut/cmd/convert.go
```diff
@@ -57,30 +67,63 @@ specify which modules should be used when generating task templates. Control-OCC
...
+			err = converter.GenerateWorkflowTemplate(role, outputDir)
 			if err != nil {
 				err = fmt.Errorf("conversion to workflow failed for %s: %w", dumpFile, err)
 				fmt.Println(err.Error())
 				os.Exit(1)
 			}
+
+			if isGitRepo {
+				fmt.Print("Press 'Enter' to view git diff...")
+				bufio.NewReader(os.Stdin).ReadBytes('\n')
```
teo, line 106: Can this be made nicer with `promptui` instead of reading an `Enter`? Something like `Would you like to view the git diff? [Y/n]` so pressing `Enter` still yields the same result, but with the option to refuse.

walnut/cmd/convert.go
```diff
@@ -57,30 +67,63 @@ specify which modules should be used when generating task templates. Control-OCC
...
+				fmt.Print(runGitCmd([]string{"diff"}))
+
+				fmt.Print("Press 'Enter' to view git status...")
```
teo, line 109: The `git status` is small anyway, so I think it should be just printed without prompting, and before the `diff`.

walnut/converter/converter.go
```diff
@@ -138,9 +138,9 @@ func ExtractTaskClasses(dplDump Dump, envModules []string) (tasks []*task.Class,
 func sanitizeCmdLineArgs (input []string, taskName string) (output []string) {
 	for index, value := range input {
 		// Check args for dump arguments and remove them
-		if  strings.Contains(value, "--dump-workflow") ||
-			strings.Contains(value, "--dump-workflow-file") ||
-			strings.Contains(value, ".json") && input[index-1] == "--dump-workflow-file" {} else {
+		if  !strings.Contains(value, "--dump-workflow") ||
+			!strings.Contains(value, "--dump-workflow-file") ||
+			!strings.Contains(value, ".json") && input[index-1] != "--dump-workflow-file" {
```
teo, line 143: Let's unpack this: you have 3 clauses separated by `||`, and the last of these has a `&&` inside (`&&` is evaluated before `||` similarly to how multiplication takes precedence over addition). You want to negate the whole if-expression to get rid of the unelegant `{} else` construct, but much like in algebra you cannot simply negate each of the operands to negate the whole clause. In other words, it is **not true** that NOT(A OR B) == NOT(A) OR NOT(B).

You could always write
```
if !(strings.Contains(value, "--dump-workflow") ||
	(value, "--dump-workflow-file") ||
	(value, ".json") && input[index-1] == "--dump-workflow-file") {
```
and this solution would be correct, but if you wish to make the expression more elegant and only do negations on the innermost statements, you need to apply De Morgan's laws.

I advise you take a look at https://en.wikipedia.org/wiki/De_Morgan%27s_laws and (protip) remember them, because negating a clause is a very common task that pops up time and again in computer science.

walnut/converter/converter.go
```diff
@@ -152,17 +152,28 @@ func sanitizeCmdLineArgs (input []string, taskName string) (output []string) {
...
 		if err != nil {
 			return fmt.Errorf("marshal failed: %v", err)
 		}
-		fileName := filepath.Join(path, SingleTask.Identifier.Name + ".yaml")
+
+		fileName := filepath.Join(path, SingleTask.Identifier.Name+".yaml")
+		if _, err := os.Stat(fileName); !os.IsNotExist(err) {
+			prompt := promptui.Select{
+				Label: fmt.Sprintf("%s already exists, overwrite?", fileName),
+				Items: []string{"yes", "no"},
```
teo, line 170: I think for binary choices you need to use `promptui.Prompt` instead of `Select`, with `IsConfirm` set to `true`. Please check this. https://godoc.org/github.com/manifoldco/promptui#Prompt

Secondly, does this prompt also react to `y/n` keys or does it expect the user to type `yes` or `no`? If so, this is not what we're looking for, all binary prompts should allow `y/n` keys followed by `Enter`, and hitting `Enter` without typing anything should select the default option (either `Y` or `N`).

walnut/converter/converter.go
```diff
@@ -172,20 +183,31 @@ func GenerateTaskTemplate(extractedTasks []*task.Class, dumpFileName string) (er
...
 	if err != nil {
 		return fmt.Errorf("error converting role to YAML: %v", err)
 	}
 
-	fileName := filepath.Join(path, input.GetName() + ".yaml")
+	fileName := filepath.Join(path, input.GetName()+".yaml")
+	if _, err := os.Stat(fileName); !os.IsNotExist(err) {
+		prompt := promptui.Select{
+			Label: fmt.Sprintf("%s already exists, overwrite?", fileName),
+			Items: []string{"yes", "no"},
```
teo, line 204: Same observation concerning `promptui` as above. The text shown to the user should be `"%s already exists, overwrite? [Y/n]"` with the default option shown uppercase. To be checked if `promptui.Prompt` appends the `[Y/n]` part automatically depending on the default value, or it needs to be added.

walnut/cmd/convert.go
teo: No, `walnut` is an interactive and noninteractive console utility, not a console GUI application, so it shouldn't clear. Compare `coconut` vs `peanut` for the difference in user experience.

walnut/converter/converter.go
obviyus: `promptui.Select` does not take text input. It provides choices, that the user can navigate and select. The default selected value is `yes` and the prompt cannot be exited without choosing a value.

obviyus: In that case, it seems to me that `promptui` is a deal-breaker:
- anything other than a `y` provided to `promptui.Prompt` [results in an error](https://github.com/manifoldco/promptui/issues/81) (???)
- even if we were to catch the error, we would then have to check the value of `result` to see if it's blank (and fall back to a default yes).

obviyus: I chose to drop `promptui` in favor of [survey/v2](https://github.com/AlecAivazis/survey) (as you originally recommended).

walnut/converter/converter.go
obviyus: Understood. Apologies for the oversight. I ended up with:
```go
        if !(strings.Contains(value, "--dump-workflow") ||
            (strings.Contains(value, ".json") && input[index-1] == "--dump-workflow-file")) {
```

After some deliberation, I noticed we need not check for `--dump-workflow-file` as that should be handled by the `--dump-workflow` check. I also parenthesized the second check to make sure the `&&` executes when we need it to.

PR #141, opened 2020-07-14
[executor] Concurrent Kill and SIGTERM

miltalex:
Related to OCTRL-263 and OCTRL-262

### DIFF
--- a/executor/executable/controllabletask.go
+++ b/executor/executable/controllabletask.go
@@ -434,15 +434,32 @@ func (t *ControllableTask) Kill() error {
 		t.pendingFinalTaskStateCh <- mesos.TASK_KILLED
 	}
 
-	// TODO: do a SIGTERM before the SIGKILL
-
+	killErrCh := make(chan error)
+	var killErr error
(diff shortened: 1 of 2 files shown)


executor/executable/controllabletask.go
```diff
@@ -434,15 +434,32 @@ func (t *ControllableTask) Kill() error {
...
+				Warning("could not kill task")
+			return
+		}
+		killErrCh <- killErr
+	}()
+	
+	// Set a small timeout to SIGTERM if SIGTERM fails or timeout passes,
+	// we perform a SIGKILL.
+	select {
+	case killErr = <- killErrCh:
```
teo, line 455: In Go, `case` statements do not fall through, and unlike `switch`, `select` does not support the `fallthrough` keyword. So the `SIGKILL` isn't happening here if the `SIGTERM` errored out.

miltalex: Since we exit the goroutine before sending the error to the channel, shouldn't the timeout trigger the `SIGKILL`?

PR #149, opened 2020-07-23
Export Task PID to API

miltalex:
Related to OCTRL-72.


core/workflow/iteratorrole.go
```diff
@@ -271,3 +271,11 @@ func (i *iteratorRole) SetRuntimeVars(kv map[string]string) {
...
 	i.template.SetRuntimeVars(kv)
 }
+
+func (i *iteratorRole) GetTaskPID() string {
+	if i == nil {
+		return ""
+	}
+	pid, _ := i.template.GetUserVars().Get("taskPID")
+	return pid
+}
```
teo, line 281: I'm not sure about this. An `iteratorRole` aggregates multiple tasks, it cannot have a PID.
Shouldn't `GetTaskPID()` only be a method on `taskRole`?

core/serverutil.go
```diff
@@ -103,11 +103,13 @@ func taskToShortTaskInfo(t *task.Task) (sti *pb.ShortTaskInfo) {
 			OfferId: t.GetOfferId(),
 			ExecutorId: t.GetExecutorId(),
 		},
+		Pid: "UNKNOWN",
 	}
 	parentRole, ok := t.GetParentRole().(workflow.Role)
```
teo, line 108: I suspect that if you type-assert this one to `taskRole` instead of `workflow.Role`, you can have `GetTaskPID()` only in `taskRole`. Alternatively, perhaps `Pid` shouldn't be a member of any role at all, it could come directly from the `task.Task`. Intuitively this makes sense for me, because a PID is not a variable associated with a logical role, it is a fundamental trait of a running `Task`.

core/workflow/rolebase.go
```diff
@@ -350,4 +350,12 @@ func (r *roleBase) CollectInboundChannels() (channels []channel.Inbound) {
 		channels = channel.MergeInbound(r.Bind, r.parent.CollectInboundChannels())
 	}	
 	return
+}
+
+func (r *roleBase) GetTaskPID() string {
```
teo, line 355: Related to the other comments, perhaps `GetTaskPID` could only be a member of `taskRole`, or alternatively, it could be outside the role system entirely and come directly from `task.Task`.

core/scheduler.go
```diff
@@ -331,6 +331,17 @@ func incomingMessageHandler(state *internalState, fidStore store.Singleton) even
...
+		case "TaskMessage":
+			var taskMessage event.TaskMessageBase
+			err = json.Unmarshal(data, &taskMessage)
+			if err != nil {
+				return
+			}
+
+			t := state.taskman.GetTask(taskMessage.GetTaskId())
+			if t != nil {
+				t.SetTaskPID(taskMessage.GetTaskPID())
```
teo, line 343: Just a side note - once `scheduler.go` is wrapped into `task.Manager`, this call will be internal to the `task` package so `setTaskPID` should be unexported.

PR #150, opened 2020-07-23
[walnut] verbose validation errors

obviyus:
Fixes OCTRL-294.

### DIFF
--- a/walnut/schemata/validate.go
+++ b/walnut/schemata/validate.go
@@ -57,19 +57,17 @@ func Validate(input []byte, format string) (err error) {
 	schemaLoader := gojsonschema.NewStringLoader(schema)  // load schema
 	documentLoader := gojsonschema.NewGoLoader(inputData) // load unmarhsaled YAML
 
-	// fmt.Printf("RAWYAML: %v\n", inputData.value)
-
 	result, err := gojsonschema.Validate(schemaLoader, documentLoader)
 	if err != nil {
 		return fmt.Errorf("Error loading data: %w", err)
 	}
 
 	if result.Valid() {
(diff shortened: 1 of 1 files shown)


obviyus:
When failing validation, instead of just an `exit.Code(1)`, `walnut.schemata` will now give a reason why validation failed:
```go
task validation failed: wants.cpu: Invalid type. Expected: number, given: string
```

walnut/schemata/validate.go
```diff
@@ -57,19 +57,17 @@ func Validate(input []byte, format string) (err error) {
...
 		return fmt.Errorf("Error loading data: %w", err)
 	}
 
 	if result.Valid() {
-		// fmt.Printf("\nSUCCESS! File is valid against %s schema\n", format)
 		os.Exit(0)
 	} else {
-		err = errors.New("file is invalid against schema")
-		return fmt.Errorf("schema validation: %w\n", err)
+		err = errors.New(fmt.Sprintf("%v", result.Errors()[0]))
```
teo, line 68: `fmt.Errorf` instead of `errors.New` plus `fmt.Sprintf`.

Also, you should use `%w` instead of `%v` when wrapping an error type within your new error. See https://blog.golang.org/go1.13-errors

obviyus: A couple of issues with this:
- `result.Error()` returns an array. We could wrap this but
- `result.Error()` returns an array of `ResultError`s

We could wrap this array of `ResultError`s into a single `fmt.Sprintf` but from my testing it seems `gojsonschema` returns a `ResultError` as soon as it encounters an error. i.e. `result.Error()` will return an array with one element.

PR #153, opened 2020-07-30
[walnut] load existing WFTs into walnut

obviyus:
Fixes OCTRL-290.


core/workflow/loadDPL.go
```diff
@@ -65,8 +65,27 @@ func LoadDPL(tasks []*task.Class, rootRoleName string) (workflow Role, err error
...
+	root := new(aggregatorRole)
+	err = yaml.Unmarshal(input, &root)
+	if err != nil {
+		return workflow, err
+	}
+
+	workflow = root
+
+	return workflow, nil
 }
```
obviyus, line 91: At this point, I do not feel that `loadDPL.go` should be a separate file. IMO, `load.go` is a more suitable location for it's functions. Let me know if it's alright to move its contents.

teo: Feel free to make this change in a new PR

PR #154, opened 2020-07-30
Support for Hooks

teo:
OCTRL-139
OCTRL-141
OCTRL-303

### DIFF
--- a/core/task/task.go
+++ b/core/task/task.go
@@ -35,6 +35,7 @@ import (
 	"strconv"
 	"strings"
 	texttemplate "text/template"
+	"time"
 
 	"github.com/AliceO2Group/Control/common"
 	"github.com/AliceO2Group/Control/common/controlmode"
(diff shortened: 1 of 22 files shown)


core/task/task.go
```diff
@@ -57,6 +58,7 @@ type parentRole interface {
 	UpdateState(State)
 	GetPath() string
 	GetTaskClass() string
+    GetTaskTraits() Traits
```
miltalex, line 61: There seems to be an indentation issue here.

core/workflow/aggregator.go
```diff
@@ -132,6 +132,18 @@ func (r *aggregator) GetTasks() (tasks task.Tasks) {
 	return
 }
 
+func (r * aggregator) GetHooksForTrigger(trigger string) (tasks task.Tasks) {
```
miltalex, line 135: There is a space here between pointer and `aggregator`.

PR #159, opened 2020-08-10
[walnut] add support for workflow grafting

obviyus:
Fixes OCTRL-281.


core/workflow/graft.go
```diff
@@ -0,0 +1,80 @@
...
+
+package workflow
+
+import (
+    "fmt"
+    "gopkg.in/yaml.v3"
+    "strings"
+)
+
+var parent *yaml.Node
```
obviyus, line 33: Not a huge fan of having this stray global variable here, but I cannot think of a way to avoid it either.

PR #167, opened 2020-08-23
[walnut] graft role as child of specified role

obviyus:
Fixes OCTRL-311.

### DIFF
--- a/core/workflow/graft.go
+++ b/core/workflow/graft.go
@@ -46,12 +46,14 @@ func Graft(root *yaml.Node, path string, toAdd []byte, graftedName string) (out 
         return nil, err
     }
 
-    nameNode := iterateNode(root, &parent, "name")
-    for index, node := range parent.Content {
-        // Search for the name node and change the Value field for the subsequent node
-        if node == nameNode {
-            parent.Content[index + 1].Value = graftedName
-            break
+    if graftedName != "" {
+        nameNode := iterateNode(root, &parent, "name")
+        for index, node := range parent.Content {
+            // Search for the name node and change the Value field for the subsequent node
+            if node == nameNode {
+                parent.Content[index+1].Value = graftedName
+                break
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


core/workflow/graft.go
```diff
@@ -80,3 +71,29 @@ func iterateNode(node *yaml.Node, parent **yaml.Node,identifier string) (found *
     }
     return nil
 }
+
+// appendRole checks if the yaml.Node passed has a "roles" field. If it exists, the toAdd yaml.Node will
+// be appended. If not, a "roles" field will be created and toAdd will be appended to that.
```
teo, line 76: A `roles` list should never be created during grafting. Grafting only works onto an existing `roles` list. If the original workflow template (the one on top of which we're grafting, e.g. `readout-stfb`) doesn't already have a `roles` list in the expected position, we should throw error, because it means that either the workflow template is wrong, or the path expression is wrong.

In other words, we can only add a child role to an aggregator role, otherwise the operation has no meaning. Think about it: finding a `roles` means that we found an aggregator role, that is a role that represents its children. If we reached target in the document that doesn't have a `roles`, means it must be a task role or another syntactically unexpected condition. Since (1) we cannot add a child to a task role, by definition, and (2) we should not create a `roles` list in a syntactically unexpected condition because we risk ruining the document in a way that the user clearly didn't expect, we should definitely error out in this case.

teo (changes requested):

core/workflow/graft.go
```diff
@@ -31,27 +31,22 @@ import (
...
     }
 
-    // Not appending to root, only to a copy of root
-    parent.Content = append(parent.Content, roleToAdd.Content[0])
+    err = appendRole(parent, toAdd)
+    if err != nil {
+        return nil, err
+    }
 
+    root.Content[0].Content[1].Value = graftedName
```
obviyus, line 49: Hard-coding the path of `name` for the root role because ideally it should always be present at the same location. We could also perform a search through the `yaml.Node` using `iterateNode`.

teo (approved):

PR #170, opened 2020-08-27
[walnut] add workflow-name and task-name-prefix flags

obviyus:
Fixes OCTRL-280.


walnut/cmd/convert.go
```diff
@@ -235,7 +237,11 @@ func init() {
 
 	convertCmd.PersistentFlags().StringVarP(&extraVars, "extra-vars", "", "",
 		"extra vars")
-	_ = viper.BindPFlag("extra-vars", rootCmd.Flags().Lookup("static-config-uri"))
+	_ = viper.BindPFlag("extra-vars", rootCmd.Flags().Lookup("extra-vars"))
+
+	convertCmd.PersistentFlags().StringVarP(&taskNamePrefix, "task-name-prefix", "", "",
+		"prefix for task name fields")
+	_ = viper.BindPFlag("task-name-prefix", rootCmd.Flags().Lookup("task-name-prefix"))
```
obviyus, line 244: I believe these are getting a bit too messy, any better way to approach these?

teo: I don't think so. They can be bound to vars and such but you still have to register the flags with `viper`.

teo (approved):

PR #171, opened 2020-08-28
New workflow conversion and validation tool

teo:


walnut/schemata/Task.json
```diff
@@ -0,0 +1,169 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://github.com/obviyus/Controlss/walnut/schemata/Task.json",
```
miltalex, line 3: Shouldn't the `$id` be an absolute-URI in a domain which we control? (e.g `https://github.com/AliceO2Group/Control/walnut/schemata/Task.json`)

teo: Ultimately I question the need of an ID at all, but if we do define one, what you propose seems correct.

walnut/schemata/Task.json
```diff
@@ -0,0 +1,169 @@
...
+                    },
+                    "rateLogging": {
+                        "title": "Rate Logging",
+                        "type": "string"
+                    }
+                },
+                "required": [
+                    "name",
+                    "type",
+                    "target"
```
miltalex, line 86: Not always present see [stfb-internal-dpl-global-binary-file-sink.yaml](https://github.com/AliceO2Group/ControlWorkflows/blob/flp-suite-v0.10.0/tasks/stfb-internal-dpl-global-binary-file-sink.yaml)

teo: You're right, `target` can be omitted.

miltalex (dismissed):

miltalex (approved):

PR #184, opened 2020-11-09
[core] fix discrepancy between task and environment states

miltalex:


core/environment/transition_configure.go
```diff
@@ -183,9 +183,11 @@ func (t ConfigureTransition) do(env *Environment) (err error) {
 		// err = t.taskman.ConfigureTasks(env.Id().Array(), tasks)
 		taskmanMessage := task.NewEnvironmentMessage(taskop.ConfigureTasks, env.Id(), tasks, nil)
 		t.taskman.MessageChannel <- taskmanMessage
-		// if err != nil {
-		// 	return
-		// }
+	}
+	incomingEv := <-env.stateChangedCh
```
teo, line 187: This is a block point... If I understand correctly we're waiting for the overall update of all task state changes relative to this environment, that follows the `ConfigureTasks` operation.

What if an update never comes?

core/environment/transition_reset.go
```diff
@@ -53,19 +53,15 @@ func (t ResetTransition) do(env *Environment) (err error) {
...
-	// 	env.Workflow().GetTasks(),
-	// 	task.CONFIGURED.String(),
-	// 	task.RESET.String(),
-	// 	task.STANDBY.String(),
-	// 	nil,
-	// )
-	// if err != nil {
-	// 	return
-	// }
+	incomingEv := <-env.stateChangedCh
```
teo, line 60: Same comment as above, do we need a select/timeout here to ensure we don't stay blocked forever?

core/environment/transition_startactivity.go
```diff
@@ -71,21 +71,17 @@ func (t StartActivityTransition) do(env *Environment) (err error) {
...
 	t.taskman.MessageChannel <- taskmanMessage
 
-	// err = t.taskman.TransitionTasks(
-	// 	env.Workflow().GetTasks(),
-	// 	task.CONFIGURED.String(),
-	// 	task.START.String(),
-	// 	task.RUNNING.String(),
-	// 	args,
-	// )
+	incomingEv := <-env.stateChangedCh
```
teo, line 78: As above.

core/environment/transition_stopactivity.go
```diff
@@ -58,20 +58,15 @@ func (t StopActivityTransition) do(env *Environment) (err error) {
...
-	// 	task.RUNNING.String(),
-	// 	task.STOP.String(),
-	// 	task.CONFIGURED.String(),
-	// 	nil,
-	// )
-
-	// if err != nil {
-	// 	return
-	// }
+	incomingEv := <-env.stateChangedCh
```
teo, line 65: ...

core/task/manager.go
```diff
@@ -783,9 +783,15 @@ func (m *Manager) handleMessage(tm *TaskmanMessage) (error) {
 	case taskop.AcquireTasks:
 		go m.acquireTasks(tm.GetEnvironmentId(), tm.GetDescriptors())
```
teo, line 784: Now that I look at it again I'm a bit concerned about this (& apologies if I'm reopening a previously settled matter, I'd just like to be doubly sure). `acquireTasks` is not mutexed from other operations, yet the fact that all of them here are dispatched as separate goroutines makes it possible for e.g. `acquireTasks` to start concurrently with `transitionTasks`.

Is the thread safety of `m.roster` enough to guarantee the overall thread safety of these concurrent calls?

teo (changes requested):

core/task/manager.go
miltalex: The `m.roster` is safe when it comes to doing operations with the roster. But with the current `transition_configure.go` it is impossible to make a `transitionTasks` call before the `acquireTasks` is finished. Cause in the `transition_configure.go` we wait for the `wfStatus` to be active, before issuing a `configureTasks` call.

teo (approved):

PR #187, opened 2020-11-12
Environment autorun feature

miltalex:


core/task/manager.go
```diff
@@ -695,14 +704,20 @@ func (m *Manager) KillTasks(taskIds []string) (killed Tasks, running Tasks, err
...
 
 	killed, running, err = m.doKillTasks(toKill)
 	for _, id := range killed.GetTaskIds() {
-		<- m.ackKilledTasks[id]
-		close(m.ackKilledTasks[id])
-		delete(m.ackKilledTasks, id)
+		ack,_ := m.ackKilledTasks.getValue(id)
+		<- ack
+		close(ack)
+		m.ackKilledTasks.deleteKey(id)
```
teo, line 715: Was there a race for the ackKilledTasks structure?

teo (changes requested):

coconut/cmd/environment_autorun.go
```diff
@@ -0,0 +1,55 @@
...
+	Long: fmt.Sprintf(`The environment create auto command requests from %s the
+creation of a new autorun environment.
+
+The operation may or may not be successful depending on available resources and configuration.`, product.PRETTY_SHORTNAME),
+	Run:   control.WrapCall(control.CreateAutoEnvironment),
+
+}
+
+func init() {
+	environmentCreateCmd.AddCommand(environmentCreateAutoCmd)
```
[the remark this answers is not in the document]
miltalex: No benefit. Actually, I think it is better with the `--auto` flag. I will change it

core/environment/manager.go
```diff
@@ -340,3 +340,67 @@ func (envs *Manager) handleDeviceEvent(evt event.DeviceEvent) {
...
+	// now we have the environment we should transition to start
+	trans := NewStartActivityTransition(envs.taskman)
+	if trans == nil {
+		// environment error should report
+		envs.taskman.MessageChannel <- task.NewEnvironmentErrorMessage(err.Error())
+		return
+	}
+
+	err = env.TryTransition(trans)
+	// this stays here as a reminder that we should remove it
```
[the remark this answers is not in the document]
miltalex: This is an old comment, I should remove it.

core/protos/coreevents.go
```diff
@@ -75,24 +76,13 @@ func NewEventTaskState(taskid,state string) *Event {
 	return WrapEvent(&stCh)
 }
 
-func NewEventTaskStatus(status *mesos.TaskStatus) *Event {
-
-	taskId := status.GetTaskID().Value
-	var statuschanged Ev_TaskStatusChanged
+func NewEventTaskStatus(taskid,status string) *Event {
```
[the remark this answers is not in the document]
miltalex: I still push the `ACTIVE`/`INACTIVE` and plus the `KILLED` status, I just moved it inside `taskman` and I push it from `updateTaskStatus` instead of the `scheduler.go`.

core/task/manager.go
miltalex: I received a `WARNING DATA RACE`, so I thought it will be safer to fix this.

coconut/control/control.go
```diff
@@ -200,6 +200,39 @@ func CreateEnvironment(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.C
 		return
 	}
 
+	if cmd.Flags().Changed("auto") {
```
teo, line 203: If I understand correctly that you're checking whether `auto` is set, `GetBool` seems more appropriate than `Changed`. The former returns the absolute value of the flag, the latter tells you whether a setting was passed that overrides any default.

core/server.go
```diff
@@ -657,25 +657,43 @@ func (m *RpcServer) SetRepoDefaultRevision(cxt context.Context, req *pb.SetRepoD
...
+		select {
+		case event, ok := <- ch:
+			if ok {
+				err := srv.Send(event)
+				if err != nil {
+					return err
+				}
+			}
+		case err := <- sub.Err():
+            return err
```
teo, line 696: Indentation issue

core/server.go
```diff
@@ -657,25 +657,43 @@ func (m *RpcServer) SetRepoDefaultRevision(cxt context.Context, req *pb.SetRepoD
...
+	ch := make(chan *pb.Event)
+	// use to subscribe to taskman
+	// here just do it
+	sub := m.state.PublicEventFeed.Subscribe(ch)
+	go m.state.environments.CreateAutoEnvironment(request.GetWorkflowTemplate(), request.GetVars(), sub)
+	for {
+		select {
+		case event, ok := <- ch:
+			if ok {
+				err := srv.Send(event)
```
teo, line 690: We still have the option of sending out Role events instead of task events, but in this case we could consider emitting Task events too with relative ease. Assuming the `publicEventFeed` gives us *all* task events of the system, we can simply check whether the event's task's environment id equals that of the current environment, and if yes, forward it. It would require some changes to `CreateAutoEnvironment` I assume, because you'd need to find out the new environment id, but it could be a reasonable way. Let's discuss.

teo (changes requested):

core/server.go
```diff
@@ -206,11 +208,11 @@ func (m *RpcServer) NewEnvironment(cxt context.Context, request *pb.NewEnvironme
...
-	select {
-	case m.state.PublicEvent <- pb.NewEnvironmentCreatedEvent(ei):
-	default:
-		log.Debug("state.PublicEvent channel is full")
-	}
+	// select {
+	// case m.state.PublicEvent <- pb.NewEnvironmentCreatedEvent(ei):
+	// default:
+	// 	log.Debug("state.PublicEvent channel is full")
+	// }
```
teo, line 215: Is this WIP?

core/server.go
```diff
@@ -377,11 +379,11 @@ func (m *RpcServer) DestroyEnvironment(cxt context.Context, req *pb.DestroyEnvir
...
-	select {
-	case m.state.PublicEvent <- pb.NewEnvironmentDestroyedEvent(ctr, env.Id().String()):
-	default:
-		log.Debug("state.PublicEvent channel is full")
-	}
+	// select {
+	// case m.state.PublicEvent <- pb.NewEnvironmentDestroyedEvent(ctr, env.Id().String()):
+	// default:
+	// 	log.Debug("state.PublicEvent channel is full")
+	// }
```
teo, line 386: Is this WIP?

core/serverutil.go
```diff
@@ -158,3 +160,34 @@ func workflowToRoleTree(root workflow.Role) (ri *pb.RoleInfo) {
 	}
 	return
 }
+
+
+type Connections struct {
```
teo, line 165: Is this part of a channel multiplexer? I feel it could use some inline docs or a more specific name.

core/task/safeacks.go
```diff
@@ -0,0 +1,79 @@
...
+ * Intergovernmental Organization or submit itself to any jurisdiction.
+ */
+
+package task
+
+import (
+	"sync"
+)
+
+type safeAcks struct {
```
teo, line 31: Same situation as `Connections`... channel leaks are pretty much the only kind of memory leak you can get in Go, so whenever we pool channels in a slice it's good to comment to explain what exactly is going on here.

teo (changes requested):

teo (approved):

PR #202, opened 2021-01-22
Fix deadlock and nil pointer crash

miltalex:


common/event/tasksreleasedevent.go
```diff
@@ -41,14 +41,23 @@ func (tr *TasksReleasedEvent) GetName() string {
 }
 
 func (tr *TasksReleasedEvent) GetEnvironmentId() uid.ID {
+	if tr == nil {
+		return ""
+	}
```
teo, line 46: This is all good & I fully agree, but were we triggering this kind of case? It might generally be better for the caller to also check whether the `TasksReleasedEvent` is `nil`, otherwise this condition stays silent. Same goes for other such funcs in this PR.

core/environment/environment.go
```diff
@@ -417,7 +414,12 @@ func (env *Environment) subscribeToWfState(taskman *task.Manager) {
...
 func (env *Environment) unsubscribeFromWfState() {
-	if env.unsubscribe != nil {
-		env.unsubscribe <- struct{}{}
+	// Use select to unblock in case the above goroutine
+	// exits due to an ERROR state. If that's the case
+	// we close the channel.
+	select {
+	case env.unsubscribe <- struct{}{}:
+	default:
+		close(env.unsubscribe)
```
teo, line 423: Perfect!

teo (approved):

PR #215, opened 2021-02-19
[core] Transition configure should return immediately after failure in deployment

miltalex:
Tested using standalone deployment.
The procedure was:

1) `coconut environment create -w readout-dataflow -e '{"hosts":["test-flp3"],"dd_enabled": "true","qcdd_enabled": "true","modulepath":"nil"}'` forcing wrong variable so that the deployment will fail
2) coconut will respond with `FATAL create:      command finished with error error=rpc error: code = DeadlineExceeded desc = context deadline exceeded`
3) After a few seconds `mesos` will inform the core about the tasks with state `TASK_FAILED` and `REASON_COMMAND_EXECUTOR_FAILED`
4) `transition_configure` will exit immediately with `workflow deployment failed`
5) clean up of the remaining tasks will start immediately

Resolves OCTRL-373

### DIFF
--- a/core/environment/transition_configure.go
+++ b/core/environment/transition_configure.go
@@ -195,6 +195,12 @@ func (t ConfigureTransition) do(env *Environment) (err error) {
 			// `environment list` to be done almost immediatelly after mesos informs with TASK_FAILED.
 			case wfState := <-notifyState:
 				if wfState == task.ERROR {
+					workflow.LeafWalk(wf, func(role workflow.Role) {
+						if st := role.GetState();  st == task.ERROR {
+							log.WithField("state", st).
+				    			Error(fmt.Sprintf("%s role failed", role.GetName()))
+						}
+					})
 					log.WithField("state", wfState.String()).
 				    	Debug("workflow state change")
 					err = errors.New("workflow deployment failed")
--- a/core/environment/transition_configure.go
+++ b/core/environment/transition_configure.go
@@ -26,6 +26,7 @@ package environment
 
 import (
 	"errors"
+	"fmt"
 	"time"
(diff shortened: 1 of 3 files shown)


core/environment/transition_configure.go
```diff
@@ -173,22 +179,33 @@ func (t ConfigureTransition) do(env *Environment) (err error) {
...
 				break WORKFLOW_ACTIVE_LOOP
+			// This is needed for when the workflow fails during the STAGING state(mesos status),mesos responds with the `REASON_COMMAND_EXECUTOR_FAILED`,
+			// By listening to workflow state ERROR we can break the loop before reaching the timeout (1m30s), we can trigger the cleanup faster
+			// in the CreateEnvironment (environment/manager.go) and the lock in the `envman` is reserved for a sorter period, which allows operations like
+			// `environment list` to be done almost immediatelly after mesos informs with TASK_FAILED.
+			case wfState := <-notifyState:
+				if wfState == task.ERROR {
+					log.WithField("state", wfState.String()).
+				    	Debug("workflow state change")
+					err = errors.New("workflow deployment failed")
```
teo, line 200: Can we also right in this spot query for **which** task(s) failed?
I expect that the WF state `ERROR` will be reported when the first `taskRole` (and therefore the first task) goes to `ERROR`. Can we query & print here which task (or role) it was that failed?

teo (changes requested):

core/environment/transition_configure.go
miltalex: I need to check if it is possible to retrieve the task or the `taskid`.

PR #221, opened 2021-03-02
Improvements to executor controllabletask.Kill

teo:
I stumbled onto this almost by accident while debugging the Kill code path for other reasons, and I suspect it really *might* fix OCTRL-401 (but I'm not sure at all, testing needed).

If my assumption is correct, I suspect the problem was that the PID used by the Kill code depends on `GetState`, and prior to this fix a failed `GetState` could result in a negative PID, which was always treated as "process not running" by the `pidExists` check. Even if this doesn't fix OCTRL-401, I think it addresses a real problem.

@miltalex I know this is assigned to you and `IN PROGRESS`, I decided to just push my fix attempt here anyway because I couldn't find a branch on your end.

### DIFF
--- a/executor/executable/controllabletask.go
+++ b/executor/executable/controllabletask.go
@@ -380,7 +380,9 @@ func (t *ControllableTask) Kill() error {
 		pid = 0
 		reachedState = "UNKNOWN" // FIXME: should be LAUNCHING or similar
 	)
-	response, err := t.rpc.GetState(context.TODO(), &pb.GetStateRequest{}, grpc.EmptyCallOption{})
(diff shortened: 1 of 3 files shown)


miltalex:
Thanks, I will give it a try tomorrow. I didn't have a branch cause I was mostly testing if the kill calls were done correctly.

miltalex (approved):
I can confirm that this was the issue for `OCTRL-401` and this works as expected.

teo:
> I can confirm that this was the issue for `OCTRL-401` and this works as expected.

Awesome! Thanks for testing @miltalex

PR #230, opened 2021-03-18
Trivial configuration endpoint for Apricot

claireguyot:
Implementation of a Trivial Configuration Endpoint (TCE) for Apricot.
* The FLPs can be retrieved with `inventory/flps`.
* The FLPs of a specific detector can be retrieved with `inventory/detectors/your_detector/flps`.
* Both can be retrieved in plain text or JSON.


apricot/local/service.go
```diff
@@ -52,9 +54,96 @@ type Service struct {
 	src cfgbackend.Source
 }
 
+type machineInfo struct {
+	name string
```
teo, line 58: Please always use 4 spaces for indentation, never tabs.

apricot/local/service.go
```diff
@@ -52,9 +54,96 @@ type Service struct {
...
+}*/
+
+func ApiGetClusterInformation(w http.ResponseWriter, r *http.Request) {
+	queryParam := mux.Vars(r)
+    format := ""
+    var err error
+	format, err = queryParam["format"]
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		fmt.Fprintf(w, "Error, format is not readable.")
```
teo, line 89: Let's not error out and default to `text` mode.

apricot/local/service.go
```diff
@@ -52,9 +54,96 @@ type Service struct {
...
+	case "text":
+		w.Header().Set("Content-Type", "text/plain")
+		w.WriteHeader(http.StatusOK)
+		//TODO: write plain text
+		/* something like: ???
+		data, _ := ioutil.ReadAll(response.Body)
+		fmt.Println(string(data))
+		*/
+	default: 
+		w.WriteHeader(http.StatusBadRequest)
```
teo, line 121: `default` should be same as `text`

apricot/local/service.go
```diff
@@ -52,9 +54,96 @@ type Service struct {
...
 	var src cfgbackend.Source
 	src, err = cfgbackend.NewSource(uri)
+	router := mux.NewRouter()
+	webApi := router.PathPrefix("/inventory/flps").Subrouter()
+    webApi.HandleFunc("/{format}", ApiGetClusterInformation).Methods(http.MethodGet)
+	webApi.HandleFunc("", ApiUnhandledRequest).Methods(http.MethodPost)
+	webApi.HandleFunc("", ApiUnhandledRequest).Methods(http.MethodPut)
+	webApi.HandleFunc("", ApiUnhandledRequest).Methods(http.MethodDelete)
+	webApi.HandleFunc("", ApiRequestNotFound)
+	log.WithError(http.ListenAndServe(uri, router)).Fatal("Fatal error with Web API.")
```
teo, line 146: The selected lines are basically your constructor. I propose you pull them out into a new function `NewHttpService`, in a separate file. This function would accept a `configuration.Service`, and can be called right after `apricot/apricot.go:L48` or similar, basically on startup of `o2-apricot`, we create the gRPC service (`s := remote.NewServer(Instance())`), and then you create the HTTP service that consumes this same `Instance()` that the gRPC server already uses.

Your `NewHttpService(svc configuration.Service)` can then construct (return) your own type, for example `apricot.TrivialService` or somesuch, with a definition like `type TrivialService struct { svc configuration.Service }`, to which a function like `ApiGetClusterInformation` can attach (as receiver).

apricot/local/service.go
```diff
@@ -52,9 +54,96 @@ type Service struct {
...
+		w.WriteHeader(http.StatusInternalServerError)
+		fmt.Fprintf(w, "Error, format is not readable.")
+		return
+	}
+	switch format {
+	case "json":
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		//TODO: write JSON
+		/* where to get info? depending on answer, could proceed like:
```
teo, line 97: See my comment concerning the constructor below. Now that you have an instance (or otherwise a pointer to the Apricot service instance) your function can become `func (ts *TrivialService) ApiGetClusterInformation(w, r)`, and then in the function body you can use `ts.svc` to acquire and then format the inventory.

teo (changes requested):

teo (changes requested):

teo (changes requested):

apricot/local/service.go
```diff
@@ -122,6 +122,23 @@ func (s *Service) GetDefaults() map[string]string {
 	return smap
 }
 
+func (s *Service) GetHostInventory(keyPrefix string) (hosts []string, err error) {
+	keys, err := s.src.GetKeysByPrefix(keyPrefix)
+	if err != nil {
+		log.WithError(err).Fatal("Error, could not retrieve host list.")
+		return []string{""}, err
```
teo, line 129: Please note that `[]string{""} != []string{}`.

The first one is a slice of strings with `len=1`, that contains a string with `len=0`. It is not an empty slice, so not quite a null value in this context.
The second one is a slice of strings with `len=0`, so an actual empty slice. This is what you should return.

apricot/local/service.go
```diff
@@ -122,6 +122,23 @@ func (s *Service) GetDefaults() map[string]string {
...
+	}
+	i := 0
+	hosts = make([]string, len(keys))
+	for _, key := range keys {
+		hostTrimed := strings.TrimPrefix(key, keyPrefix)
```
teo, line 134: `hostTrimmed`

apricot/local/servicehttp.go
```diff
@@ -0,0 +1,109 @@
...
+		Addr:         ":47188",
+		WriteTimeout: 15 * time.Second,
+		ReadTimeout:  15 * time.Second,
+	}
+	go func() {
+		apiFlps := router.PathPrefix("/inventory/flps").Subrouter()
+		apiFlps.HandleFunc("", httpsvc.ApiGetFlps).Methods(http.MethodGet)
+		apiFlps.HandleFunc("/", httpsvc.ApiGetFlps).Methods(http.MethodGet)
+		apiFlps.HandleFunc("/{format}", httpsvc.ApiGetFlps).Methods(http.MethodGet)
+		apiDetectorFlps := router.PathPrefix("/inventory/detectors/TST/flps").Subrouter()
```
teo, line 102: `TST` here should be a variable. The allowed values are in `common/systemid/systemid.go`, and they are an enum, so no strings for now. You should not use the enums, this is just to give you an idea of what could be found as input instead of `TST`.

For example the route handler for `/inventory/detectors/ITS/flps` should return whatever's to be found in `o2/hardware/detectors/ITS/flps/`.

apricot/local/servicehttp.go
```diff
@@ -0,0 +1,109 @@
...
+	svc configuration.Service
+}
+
+func (httpsvc *HttpService) ApiGetFlps(w http.ResponseWriter, r *http.Request) {
+	keyPrefix := "o2/hardware/flps/"
+	httpsvc.ApiGetClusterInformation(w, r, keyPrefix)
+}
+
+func (httpsvc *HttpService) ApiGetDetectorFlps(w http.ResponseWriter, r *http.Request) {
+	keyPrefix := "o2/hardware/detectors/TST/flps/"
```
teo, line 47: `TST` here should be a variable. See my other comment in `NewHttpService` below.

apricot/local/servicehttp.go
```diff
@@ -0,0 +1,109 @@
...
+func (httpsvc *HttpService) ApiGetClusterInformation(w http.ResponseWriter, r *http.Request, keyPrefix string) {
+	queryParam := mux.Vars(r)
+	format := ""
+	format = queryParam["format"]
+	if format == "" {
+		format = "text"
+	}
+	keys, err := httpsvc.svc.GetHostInventory(keyPrefix)
+	if err != nil {
+		log.WithError(err).Fatal("Error, could not retrieve host list.")
```
teo, line 60: Not really a `Fatal` IMO, `Fatal` should only be used if we're forced to shut down the instance or the instance is already dying/crashing. For me this is an honest `Warn`.

apricot/local/servicehttp.go
```diff
@@ -0,0 +1,109 @@
...
+		for _, hostname := range keys {
+			fmt.Fprintf(w,"%s\n", hostname)
+		}
+	default:
+		w.Header().Set("Content-Type", "text/plain")
+		w.WriteHeader(http.StatusOK)
+		for _, hostname := range keys {
+			fmt.Fprintf(w,"%s\n", hostname)
+		}
+	}
```
teo, line 83: This is all very good!

Remarks:
  * Please `Fatal`→`Warn`
  * The code for the `"text"` and `default` cases is identical, so you can do:
```
	case "text": fallthrough
	default:
		w.Header().Set("Content-Type", "text/plain")
		w.WriteHeader(http.StatusOK)
		for _, hostname := range keys {
			fmt.Fprintf(w,"%s\n", hostname)
		}
```
... or even skip the `"text"` case altogether, as there are only 2 possible behaviours.

apricot/local/servicehttp.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func NewHttpService(service configuration.Service) (svr *http.Server) {
+	router := mux.NewRouter()
+	httpsvc := &HttpService{
+		svc: service,
+	}
+	httpsvr := &http.Server{
+		Handler:      router,
+		Addr:         ":47188",
```
teo, line 93: Please look into `apricot/config.go`, you'll find a bunch of `viper.SetDefault` and `pflag.*` calls. The purpose of these is respectively to set a fallback value associated with a string label, and to fetch a value from command line arguments and parse/associate it with a string label.

I propose that you define a new flag `httpListenPort` similar to the existing `listenPort`, and then use it here.

apricot/local/servicehttp.go
```diff
@@ -0,0 +1,109 @@
...
+		WriteTimeout: 15 * time.Second,
+		ReadTimeout:  15 * time.Second,
+	}
+	go func() {
```
teo, line 97: Since the listener is separate from the `HttpService` now, I'm not sure you need the `go func` here any more. Have you tried running the content of this closure simply in the constructor?

apricot/protos/apricot.proto
```diff
@@ -131,6 +132,14 @@ message ComponentEntriesResponse {
     repeated string payload = 1;
 }
 
+message HostGetRequest {
+    string prefix = 1;
```
teo, line 136: I'd propose an alternative structure here. The `keyPrefix` is an important piece of information when you query Consul as your data backend, but we should not expose it to the user. In other words, the user knows whether they want
1) all flps, or
2) only flps assigned to a certain `detector`.

Therefore:
```
message HostGetRequest {
    string detector = 1;
}
```

Now, in this instance we can take advantage of the fact that all Protobuf fields are optional, as well as having well-defined null values. If the `detector` string you get is empty, you return all FLPs. Otherwise, you only return those for a given detector. In both cases, the `keyPrefix` is a piece of hardcoded data-structure information that should stay within Apricot.

apricot/remote/service.go
```diff
@@ -193,6 +193,16 @@ func (c *RemoteService) SetRuntimeEntry(component string, key string, value stri
 	return nil
 }
 
+func (c *RemoteService) GetHostInventory(keyPrefix string) (hosts []string, err error) {
```
teo, line 196: See my comment on `keyPrefix` as parameter vs. `detector` as parameter.

apricot/signals.go
```diff
@@ -50,6 +52,9 @@ func signals(srv *grpc.Server) {
 			Debug("received signal")
 
 		srv.Stop()
+		if err := httpsvr.Shutdown(context.Background()); err != nil {
+			log.Printf("Error while shutting down http server.")
```
teo, line 56: `log.Warn`

configuration/service.go
```diff
@@ -41,6 +41,7 @@ type Service interface {
 	GetComponentConfiguration(query *componentcfg.Query) (payload string, err error)
 	GetAndProcessComponentConfiguration(query *componentcfg.Query, varStack map[string]string) (payload string, err error)
 
+	GetHostInventory(keyPrefix string) (hosts []string, err error)
```
teo, line 44: I propose `GetHostInventory(detector string) (hosts []string, err error)` as previously mentioned.

apricot/apricot.go
```diff
@@ -28,6 +28,7 @@ import (
 	"fmt"
 	"net"
 
+    "github.com/AliceO2Group/Control/apricot/local"
```
teo, line 31: There should be a tab here.

apricot/local/service.go
```diff
@@ -122,6 +122,23 @@ func (s *Service) GetDefaults() map[string]string {
 	return smap
 }
 
+func (s *Service) GetHostInventory(keyPrefix string) (hosts []string, err error) {
```
teo, line 125: Please see my comment further below on `keyPrefix` vs. `detector` as parameter.

I think `keyPrefix` should be a constant `inventoryKeyPrefix` defined at the top of this file, and then if you receive a non-empty `detector`, you amend the prefix you pass to `GetKeysByPrefix` as needed.

teo (changes requested):

teo (changes requested):
No further objections on the code, but the `inventory/flps` path returns double entries.

apricot/local/service.go
```diff
@@ -135,12 +135,18 @@ func (s *Service) GetHostInventory(detector string) (hosts []string, err error)
 		log.WithError(err).Fatal("Error, could not retrieve host list.")
 		return []string{}, err
 	}
-	i := 0
-	hosts = make([]string, len(keys))
+	hostSet := make(map[string]bool, len(keys))
```
teo, line 138: I advise starting from zero length here, because `len(keys)` is only an upper bound here.

apricot/local/service.go
```diff
@@ -135,12 +135,18 @@ func (s *Service) GetHostInventory(detector string) (hosts []string, err error)
...
 	for _, key := range keys {
 		hostTrimmed := strings.TrimPrefix(key, keyPrefix)
 		hostname := strings.Split(hostTrimmed, "/")
-		hosts[i] = hostname[0]
+		if _, ok := hostSet[hostname[0]]; !ok {
+			hostSet[hostname[0]] = true
+		}
```
teo, line 144: Since this is a set we can assume a write is fast. Better to just write directly i.e. `hostSet[hostname[0]] = true`. Worst case you're gonna overwrite entries with identical ones all the time, I'd argue it's unlikely that this is more expensive than checking before every write.

teo (changes requested):

teo (approved):

PR #242, opened 2021-04-14
DD scheduler polling for CONFIGURED/TERMINATED partition state

teo:
This is necessary to make DD calls blocking

### DIFF
--- a/core/integration/ddsched/plugin.go
+++ b/core/integration/ddsched/plugin.go
@@ -177,11 +177,30 @@ func (p *Plugin) ObjectStack(data interface{}) (stack map[string]interface{}) {
 				Error("failed to perform DD scheduler PartitionInitialize")
 			return
 		}
-		if response.PartitionState != ddpb.PartitionState_PARTITION_CONFIGURING {
+		if response.PartitionState != ddpb.PartitionState_PARTITION_CONFIGURING &&
+			response.PartitionState != ddpb.PartitionState_PARTITION_CONFIGURED {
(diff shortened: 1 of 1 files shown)


core/integration/ddsched/plugin.go
```diff
@@ -177,11 +177,30 @@ func (p *Plugin) ObjectStack(data interface{}) (stack map[string]interface{}) {
...
+			case ddpb.PartitionState_PARTITION_CONFIGURING:
+				time.Sleep(100 * time.Millisecond)
+			case ddpb.PartitionState_PARTITION_CONFIGURED:
+				break PARTITION_STATE_POLLING
+			default:
+				log.WithError(fmt.Errorf("PartitionInitialize landed on unexpected state %s (expected: PARTITION_CONFIGURED)", response.PartitionState.String())).
+					WithField("environment_id", envId).
+					WithField("endpoint", viper.GetString("ddSchedulerEndpoint")).
+					Error("failed to perform DD scheduler PartitionInitialize")
+				break PARTITION_STATE_POLLING
```
teo, line 202: @ironMann is this the polling logic you had in mind for DD sched state checking?

ironMann: I'm not sure what this code does, but Start button (start transition) should not be enabled until DD returns `PARTITION_CONFIGURED`.  You can log the message filed of that response to have more info about what is happening with the partition, while it is in the transition state (`State_PARTITION_CONFIGURING`)

PR #245, opened 2021-04-21
[repos] Only list public workflows by default

kostorr:
Closes OCTRL-458, OCTRL-466


core/repos/repo.go
```diff
@@ -113,17 +102,20 @@ func (r *Repo) getCloneParentDirs() []string {
...
 		r.HostingSite + "/" +
-		r.User 		  + "/" +
-		r.RepoName	  + ".git"
+		r.User + "/" +
+		r.RepoName + ".git"
 }
 
 func (r *Repo) getWorkflowDir() string {
-	return r.getCloneDir() + "/workflows/"
+	return filepath.Join(r.getCloneDir(), "workflows")
```
teo, line 110: Note that here the behavior changes because `filepath.Join` doesn't append a trailing slash. You should check whether `getWorkflowDir()` is called in contexts where a trailing slash is expected.

core/repos/repo.go
```diff
@@ -113,17 +102,20 @@ func (r *Repo) getCloneParentDirs() []string {
...
 }
 
 func (r *Repo) ResolveTaskClassIdentifier(loadTaskClass string) (taskClassIdentifier string) {
 	if !strings.Contains(loadTaskClass, "/") {
-		taskClassIdentifier = r.HostingSite + "/" + r.User + "/" + r.RepoName + "/tasks/" + loadTaskClass
+		taskClassIdentifier = filepath.Join(r.HostingSite,
```
teo, line 115: This is not equivalent, you're missing `/tasks/` from the path.

core/repos/repo.go
```diff
@@ -231,14 +227,21 @@ func (r *Repo) getWorkflows(revisionPattern string, gitRefs []string) (Templates
...
 		}
 		for _, file := range files {
-			if strings.HasSuffix(file.Name(), ".yaml") { // Only return .yaml files
-				templates[RevisionKey(revision)] = append(templates[RevisionKey(revision)], Template(strings.TrimSuffix(file.Name(), ".yaml")))
+			// Only return .yaml files
+			if strings.HasSuffix(file.Name(), ".yaml") {
+				// Check if workflow is public in case not allWorkflows requested
+				// and skip it if it isn't
+				if allWorkflows || (!allWorkflows &&
+					utils.IsFilePublicWorkflow(filepath.Join(r.getWorkflowDir(), file.Name()))) {
```
teo, line 235: Good, this is exactly what I had in mind.

core/repos/repomanager.go
```diff
@@ -148,11 +159,13 @@ func (manager *RepoManager)  discoverRepos() (repos []string, err error){
...
 
 			for _, repo := range someRepos { //sanitize path
+				// Get rid of invalid paths
+				if !isValidPath(repo) { continue }
 				repoDir := manager.rService.GetReposPath()
-				utils.EnsureTrailingSlash(&repoDir)
-				repo = strings.TrimPrefix(repo, repoDir)
-				utils.EnsureTrailingSlash(&repo)
-				repos = append(repos, repo)
+				repo, err = filepath.Rel(repoDir, repo) // trim repoDir
```
teo, line 165: I wonder if you still need to `EnsureTrailingSlash` after this. I'm not sure, but please double check that the behavior doesn't change.

I see you remove the same call in other spots too. In principle that's ok as long as we ensure it's all internally consistent.

core/repos/repomanager.go
```diff
@@ -454,7 +463,7 @@ func (manager *RepoManager) GetWorkflow(workflowPath string)  (resolvedWorkflowP
 	if !strings.HasSuffix(workflowFile, ".yaml") { //Add trailing ".yaml"
 		workflowFile += ".yaml"
 	}
-	resolvedWorkflowPath = workflowRepo.getWorkflowDir() + workflowFile
+	resolvedWorkflowPath = filepath.Join(workflowRepo.getWorkflowDir(), workflowFile)
```
teo, line 466: Thanks to the fact that you use `filepath.Join`, this will work regardless of whether `getWorkflowDir` returns a trailing slash or not. But other context of `getWorkflowDir` should be checked too.

teo (changes requested):
My comments are mostly minor, and based on the fact that modifying trailing slash behavior triggers my fear (paranoia?) of introducing silent bugs.

Except for a few minor spots where we clearly have an issue (e.g. the `tasks` path building), this can in principle be merged, as long as you've ensured consistency between all contexts where the trailing slash behavior changes.

kostorr:
> My comments are mostly minor, and based on the fact that modifying trailing slash behavior triggers my fear (paranoia?) of introducing silent bugs.
> 
> Except for a few minor spots where we clearly have an issue (e.g. the `tasks` path building), this can in principle be merged, as long as you've ensured consistency between all contexts where the trailing slash behavior changes.

I share the paranoia which is why I thought better to move to join altogether. I was going through the code and it was impossible to remember why the slashes were needed here and there. 

There were indeed some loose ends which I addressed. Creating and starting environments everything seems to be working fine.

teo (approved):

PR #249, opened 2021-04-26
Expose workflow variable information

kostorr:
Closes OCTRL-231


kostorr:
@teo this is now ready to be reviewed.

I know @graduta has some comments wrt variable structure and the corresponding protobuf messages. Any resulting changes should be very quick to implement.

teo:
@graduta should we go ahead or would you like to propose any last minute changes? I'm especially unsure about the `UiWidgets` enum: these are hints for you, so let us know whether any widget names should be added/removed/changed.

@kostorr For me this is good to merge as soon as @graduta gives his blessing. @kostorr could you please also look into displaying public VarSpec information in `coconut template`? Not urgent, I'm thinking `coconut template list` stays the same and `coconut template show` can be a new command that lists variable documentation for a given WFT.

teo (dismissed):

kostorr:
> @graduta should we go ahead or would you like to propose any last minute changes? I'm especially unsure about the `UiWidgets` enum: these are hints for you, so let us know whether any widget names should be added/removed/changed.
> 
> @kostorr For me this is good to merge as soon as @graduta gives his blessing. @kostorr could you please also look into displaying public VarSpec information in `coconut template`? Not urgent, I'm thinking `coconut template list` stays the same and `coconut template show` can be a new command that lists variable documentation for a given WFT.

Yes for sure. I'm opening a Jira ticket.

graduta (approved):
From an API point of view, it looks good to me

core/protos/o2control.proto
```diff
@@ -375,17 +375,17 @@ message GetWorkflowTemplatesRequest{
...
-        listBox = 2;
+        editBox = 0;    // plain string input line, can accept types number and string
+        slider = 1;     // input widget exclusively for numbers, range allowedValues[0]-[1]
+        listBox = 2;    // displays a list of items, can accept types number, string or list; if number/string ==> single selection, otherwise multiple selection allowed
         dropDownBox = 3;
         comboBox = 4;
     }
 
     enum Type {
-        bool = 0;
```
kostorr: These changes need to be followed by an update in the enum converter in `serverutils.go`. @graduta do we keep the same defaults as before? i.e. `UiWidget` defaults to `editBox` and `Type` defaults to `string`?

teo: Yes, we just discussed this. I reordered the enum to make `string` the first entry so that `string` becomes the implicit Protobuf default regardless of what the code does.

I will amend `serverutils.go`.

kostorr: ok pushed the change

PR #251, opened 2021-04-28
Integrated Services API

teo:


core/protos/o2control.proto
```diff
@@ -441,4 +444,16 @@ message SetRepoDefaultRevisionReply {
...
+    map<string, IntegratedServiceInfo> services = 1;
+}
+
+message IntegratedServiceInfo {
+    string name = 1;
+    bool enabled = 2;
+    string endpoint = 3;
+    string connectionState = 4;
+    string data = 5;
 }
```
teo, line 459: @graduta new Integrated Services API call, you should be able to use it to build mini-dashboards one for each service. For now there's only this call, support for events sent over the status channel should come soon (@miltalex ).

graduta (approved):

PR #261, opened 2021-05-21
Cache workflow templates to speed up API calls

kostorr:
This PR introduces a per-repo workflow cache that is repopulated on `repo refresh` iff a `git fetch` sees an update. This way file operations to check for the `!public` tag are not repeated unnecessarily. Initial cache population happens on core startup[1].

As a result API calls to get the workflows are sped up significantly. On the flip side, in case of a ControlWorkflows update, `repo refresh` will take some seconds (<10 with the current ControlWorkflows repo). Indicatively:

```
# v0.23.0
$ time (./coconut repo refresh > /dev/null)                                                                                                                                                                       
                                                                                                                                                                                                                                      
real    0m0.330s                                                                                                                                                                                                                      
user    0m0.025s                                                                                                                                                                                                                      
sys     0m0.029s

$ time (./coconut templ list > /dev/null)                                                                                                                                                                         
                                                                                                                                                                                                                                      
real    0m0.278s                                                                                                                                                                                                                      
user    0m0.022s                                                                                                                                                                                                                      
sys     0m0.028s

$ time (./coconut templ list --all-workflows --all-branches > /dev/null) #equivalent of GUI call                                                                                                                      
                                                                                                                                                                                                                                      
real    0m3.205s                                                                                                                                                                                                                      
user    0m0.026s                                                                                                                                                                                                                      
sys     0m0.033s
```

```
# with cache
$ time (./coconut repo refresh > /dev/null) #no git update                                                                                                                                                                       
                                                                                                                                                                                                                                      
real    0m0.353s                                                                                                                                                                                                                      
user    0m0.021s                                                                                                                                                                                                                      
sys     0m0.024s

$ time (./coconut repo refresh > /dev/null) #git update                                                                                                                                                            
                                                                                                                                                                                                                                      
real    0m5.198s                                                                                                                                                                                                                      
user    0m0.029s                                                                                                                                                                                                                      
sys     0m0.040s  

$ time (./coconut templ list > /dev/null)                                                                                                                                                                         
                                                                                                                                                                                                                                      
real    0m0.056s                                                                                                                                                                                                                      
user    0m0.021s                                                                                                                                                                                                                      
sys     0m0.026s     

$ time (./coconut templ list --all-branches --all-workflows > /dev/null) #equivalent of GUI call                                                                                                                                           
                                                                                                                                                                                                                                      
real    0m0.059s                                                                                                                                                                                                                      
user    0m0.020s                                                                                                                                                                                                                      
sys     0m0.027s
```

A possible improvement for the `repo refresh` time in case of a git update could be to update the cache based on the new diff instead of repopulating it.

[1] @teo @miltalex core startup on my (very weak) macbook now takes ~6s compare to ~250ms before. Will this trigger an issue along the chain?

Closes OCTRL-484.


core/core.go
```diff
@@ -68,6 +69,9 @@ func Run() error {
 	// Set up channel to receive Unix Signals.
 	signals(state)
 
+	// Start the Repo Manager instance
+	log.Infof("Starting the Control Workflows repo manager")
+	_ = the.RepoManager()
```
kostorr, line 74: @teo is this the correct place to put this?

teo (dismissed):

core/core.go
teo: Yes, I think it's a good enough place.
However 6s for startup is quite a bit, I wonder if there could be some low hanging fruit in the !public parsing that allows us to trim it down.

If you feel like going down the optimization rabbit hole, `utils.TimeTrack` and `utils.TimeTrackFunction` can help.

graduta: I agree that `6s` might be a little bit problematic especially within the Ansible recipe. Other services are dependent on `AliECS Core` being up and running. 

Could we maybe start the server and after that, in the background, Core does the reading and parsing of the files?
I believe there is no `GetWorkflowTemplates` call right at the start of the server.

kostorr:
The `repos` functionality has been updated to 
* support an on-demand updateable template cache
* shell out for a `git checkout` operation (as go-git implements a slightly different, more costly, procedure)
* ensure that the repo is configured to be case insensitive (covers a file rename between revisions, causing a failure)

In order to accommodate the needs of the GUI the `ListRepos` call has been extended to also optionally provide the revisions of a repo. There is still a small matter on the interface to be discussed with George.

core/core.go
kostorr: Slow startup addressed by the latest changes.

kostorr:
Following our discussion, and seeing how this is backwards compatible, this can now be merged.

teo (approved):

PR #264, opened 2021-05-27
[core] Add timeout to Partition state polling

miltalex:


core/integration/ddsched/plugin.go
```diff
@@ -234,9 +234,9 @@ func (p *Plugin) ObjectStack(data interface{}) (stack map[string]interface{}) {
 				Error("failed to perform DD scheduler PartitionInitialize")
 			return
 		}
-
+		pollingTimeout := 15 * time.Second
```
teo, line 237: Please read this value from a variable `dd_polling_timeout` (you can access `varStack` from this context). If the variable is not defined, we default to `30s`. The reason for this is that depending on the workflow structure, different DD timeouts can be appropriate.

core/integration/ddsched/plugin.go
```diff
@@ -250,6 +250,14 @@ func (p *Plugin) ObjectStack(data interface{}) (stack map[string]interface{}) {
 					Error("failed to perform DD scheduler PartitionInitialize")
 				break PARTITION_STATE_POLLING
 			}
+			if time.Since(startPolling) > pollingTimeout {
```
teo, line 253: This is perfect. Please do the same for the `PartitionTerminate` below.

teo (changes requested):

teo (approved):

PR #278, opened 2021-07-20
Environment filter for public/non-public

miltalex:


teo (approved):

core/protos/o2control.proto
```diff
@@ -183,9 +185,11 @@ message EnvironmentInfo {
 message NewEnvironmentRequest {
     string workflowTemplate = 1;
     map<string, string> vars = 2;
+    bool public = 3;
```
teo, line 188: @graduta this protofile change was just merged and will be released in AliECS v0.25.
Solves https://alice.its.cern.ch/jira/browse/OCTRL-178

graduta: What is the default value? Does it need a change in the GUI or is it core assigning it? I think it will be a bit problematic for the GUI to decide if it should be public or not?

I was under the impression that the change would be somewhere in the template definition and than the GUI would simply request all environments and specify if `public` or `non-public`.

Moreover, I believe the case in the ticket described was for `NewAutoEnvironment` (users  clicking `Clean Resources`) and not only `NewEnvironment`.

I probably lack a little bit of information here. We should probably meet and discuss a little bit more into detail.

miltalex: The default value is the same as the rpc default value for booleans which is `false`. 

With the current code/implementation, the user can only overwrite a `NewEnvironment` and make it public.
```
if request.Public {
   newEnv.Public = request.Public
}
```

So if nothing is specified from the user/API it will default to the `!public` tag from the workflows. For example, with the current implementation, all the `NewAutoEnvironment` requests will be non-public and there isn't a way to make them public from the API since the option is not provided. Running `Cleaning Resources` will create a non-public environment and the `GetEnvironments` shouldn't return it, unlesss `showAll: true`

I could change the logic, as suggested by Teo, to an `enum` with `nul`, `false`, and `true` and allow overwrite of the env's publicity to `false` .

graduta: Thank you for the detailed answer. I believe for the `AutoEnvironments` we are ok with them being not public (at least from GUI perspective) because we use them only for:
* roc-config
* cleanup-resources

While for the normal environments, just to confirm, does the GUI have to specify that when the user clicks `Create`, the GUI has to send the `public: true` option?

miltalex: For the normal environments, it shouldn't require the `public: true` only if someone wants to enforce an environment to public.

Since the GUI displays only public workflows, `core` will create public environments for those workflows without needing the `public: true`. 

P.S: The only issue with the current implementation is that the user can't overwrite a public workflow to a non-public environment using `public: false`.

graduta: Ah, ok. This is what I was not understanding. So the template  definition will be taken into consideration.
Good to know

PR #281, opened 2021-07-28
[apricot] Add ListRuntimeEntries to runtime key-value API

teo:


teo:
@graduta this unblocks progress on saving/loading parameter sets https://alice.its.cern.ch/jira/browse/OGUI-915

PR #312, opened 2021-11-18
JIT DPL workflow generation

kostorr:
Closes OCTRL-419


apricot/local/service.go
```diff
@@ -242,6 +242,45 @@ func (s *Service) GetComponentConfiguration(query *componentcfg.Query) (payload
 	return
 }
 
+func (s *Service) GetComponentConfigurationWithLastIndex(query *componentcfg.Query) (payload string, lastIndex uint64, err error) {
```
teo, line 245: Not urgent but perhaps `GetComponentConfiguration` above could internally call `GetComponentConfigurationWithLastIndex` to avoid code duplication.

configuration/template/stack.go
```diff
@@ -128,6 +135,88 @@ func MakeConfigAccessFuncs(confSvc ConfigurationService, varStack map[string]str
...
+				}
+
+				var err error
+
+				// Generate a hash out of the concatenation of
+				// 1) The full DPL command
+				// 2) The LastIndex of each payload
+				hash := sha1.New()
+				hash.Write([]byte(dplCommand + metadata))
+				jitWorkflowName := "jit-" + hex.EncodeToString(hash.Sum(nil))
```
teo, line 176: Just a note for later: in order to avoid polluting the repo dir with JIT-generated files but still maintain filesystem semantics, we could output to a subdirectory in `/tmp` (which is usually a RAM mount anyway) and then use https://github.com/dschmidt/go-layerfs to keep using the workflow loader same as before. Just an idea...

teo (approved):

PR #317, opened 2022-01-25
Reduce logging verbosity

PierreVandeVyvre:
These are the changes to reduce the verbosity of the 30 logging messages generating the largest Infologger traffic.


teo (approved):
@PierreVandeVyvre very good, thanks, I'm merging this now.

A request for your future efforts: could you please disable "format on save" in your editor? The entry should be called `editor.formatOnSave`, in your editor settings. This way we avoid generating whitespace changes in commits with the automatic reformat of the whole file every time a single line edit is saved.

PR #319, opened 2022-02-01
[core] Update PDP script call generator

teo:


teo:
@davidrohr FYI

davidrohr:
@teo @vascobarroso : I have one question: with these changes in the core, as requested, ECS will transfer the configured default versions in case the string `default` is entered in the AliECS GUI. Can you let me know where these can be configured. Also, we have to initialize it with correct default values before we test it, right?

davidrohr:
@teo : the default versions to be put in consul for now should be:
´pdp_o2_data_processing_hash` = `nightly-20220201`
`pdp_o2pdpsuite_version` = `O2PDPSuite/epn-20220203-1`
`pdp_qcjson_version` = `1`

PR #323, opened 2022-02-14
[OCTRL-122] Allow the user to set their own environment name

claireguyot:


coconut/control/control.go
```diff
@@ -392,6 +393,16 @@ func stringMapToString(stringMap map[string]string, indent string) string {
 	return strings.Join(accumulator, "\n")
 }
 
+func SetEnvironmentName(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.Command, args []string, o io.Writer) error {
+	if len(args) == 1 { // Set global default
+		fmt.Fprintln(o, "Operation failed.")
+		return fmt.Errorf("Missing Argument")
+	} else if len(args) == 2 { // Set per-repo default
+		_, _ = rpc.SetEnvironmentName(cxt, &pb.SetEnvironmentNameRequest{Name: args[0], Id: args[1]}, grpc.EmptyCallOption{})
```
teo, line 401: `SetEnvironmentName` should be allowed to fail, i.e. I advise you capture the error value and report it to the user (why the name couldn't be set), and on the server side you implement the failure modes for this call, returning any error conditions using the gRPC status idiom as you already do.

core/environment/environment.go
```diff
@@ -105,6 +107,14 @@ func newEnvironment(userVars map[string]string) (env *Environment, err error) {
 		callsPendingAwait: make(map[string]callable.CallsMap),
 	}
 
+	if envName, ok := userVars["environmentName"]; ok {
```
teo, line 110: Variables within the environment template context should use `snake_case`, so `environment_name`.

core/environment/environment.go
```diff
@@ -519,6 +529,34 @@ func (env *Environment) Id() uid.ID {
...
+	return env.name
+}
+
+func (env *Environment) SetName(name string) {
+	if env == nil {
+		return
+	}
+	env.Mu.Lock()
+	defer env.Mu.Unlock()
+	env.name = name
```
teo, line 547: The interface is correct and well designed, but I think the implementation misses a detail.

I count 2 occurrences of the environment name: the `env.name` member variable, and the `environment_name` `UserVars` entry, which ends up in the common variables stack of the template system for the given environment.

These two must be kept in sync: your `GetName` may only read `env.name`, but your `SetName` must also write into `env.UserVars`, otherwise a subsequent query for the env name from within the environment template context will yield inconsistent results.

core/environment/environment.go
```diff
@@ -519,6 +529,34 @@ func (env *Environment) Id() uid.ID {
...
+}
+
+func generateEnvironmentName() (name string) {
+	n := 5
+	b := make([]byte, n)
+	if _, err := rand.Read(b); err != nil {
+		panic(err)
+	}
+	name = fmt.Sprintf("%X", b)
+	return
```
teo, line 557: Understood, we will discuss this.

core/integration/dcs/protos/dcs.pb.go
```diff
@@ -25,7 +25,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.25.0
-// 	protoc        v3.17.3
+// 	protoc        v3.6.1
```
teo, line 28: Can you exclude this file as well as `ddsched.pb.go` from the PR? The line change was triggered by `make generate` but we don't really need to update this generated code as we don't touch anything near it.

core/integration/odc/protos/odc.proto
```diff
@@ -1,236 +1 @@
...
-//
-// Generic requests
-//
-
-// Status request
-message StatusRequest
-{
-    bool running = 1; // Select only running DDS sessions
-}
-
```
teo, line 236: Similar comment to the generated `pb.go` files, this one isn't generated but it's fetched from upstream. Either way no code changes should happen to this file, please exclude it from the PR. Let me know if you need help with this.

core/protos/o2control.proto
```diff
@@ -72,6 +72,7 @@ service Control {
     rpc ControlEnvironment (ControlEnvironmentRequest) returns (ControlEnvironmentReply) {}
     rpc ModifyEnvironment (ModifyEnvironmentRequest) returns (ModifyEnvironmentReply) {}
     rpc DestroyEnvironment (DestroyEnvironmentRequest) returns (DestroyEnvironmentReply) {}
+    rpc SetEnvironmentName (SetEnvironmentNameRequest) returns (Empty) {}
```
teo, line 75: This is all good.

core/server.go
```diff
@@ -437,6 +440,29 @@ func (m *RpcServer) DestroyEnvironment(cxt context.Context, req *pb.DestroyEnvir
...
+		return nil, status.New(codes.InvalidArgument, "received bad environment id").Err()
+	}
+
+	var env *environment.Environment
+	env, err = m.state.environments.Environment(envId)
+	if err != nil {
+		return nil, status.Newf(codes.NotFound, "environment not found: %s", err.Error()).Err()
+	}
+
+	env.SetName(req.GetName())
```
teo, line 462: The first 3 failure modes here are correct and I 100% agree with them, but we are missing a 4th one: if the user tries to set an environment name that already identifies another environment. We must do this in order to guarantee the uniqueness of env names, and since a name is not a UUID, we don't get the guarantee of uniqueness for free, and we must implement it ourselves.

I propose that the `env.SetName` function is made to return an error, and then you can forward this error by returning a gRPC `status` here. Inside `SetName`, you'd then iterate over all envs to check their names before going ahead with the write.

Depending on how you implement this, you might need to either jump one level up (for example a `environment.manager.SetEnvName(envId, newName)`) or pass a handle to the envMan to the existing function (e.g. `env.SetName(envMan, req.GetName())`). Let me know if you'd like us to discuss this further.

teo (changes requested):

teo (changes requested):

PR #329, opened 2022-03-08
[OCTRL-161] Allow for a description for workflow templates, that can be displayed in the GUI

claireguyot:


coconut/control/control.go
```diff
@@ -842,6 +845,7 @@ func ListWorkflowTemplates(cxt context.Context, rpc *coconut.RpcClient, cmd *cob
 				prevRevision = tmpl.GetRevision()
 			}
 			revBranch.AddNode(tmpl.GetTemplate())
+			revBranch.AddNode(tmpl.GetDescription())
```
teo, line 848: How is this displayed? Can you attach a screenshot?

core/protos/o2control.proto
```diff
@@ -186,6 +186,7 @@ message EnvironmentInfo {
     map<string, string> userVars = 9;
     int32 numberOfFlps = 10;
     repeated string includedDetectors = 11;
+    string rootRoleDescription = 12;
```
teo, line 189: I'd just call it `description`. We only read the description from the root role anyway, the field is ignored in other places.

core/protos/o2control.proto
```diff
@@ -429,6 +431,7 @@ message WorkflowTemplateInfo {
     string template = 2;
     string revision = 3;
     map<string, VarSpecMessage> varSpecMap = 4;
+    string description = 5;
```
teo, line 434: Oooooook, now I see what you did. You added a description field to any and all roles, not just to the root role as a way to input an environment description. This is not quite what I expected but on the other hand it might not hurt. Let's discuss this tomorrow while looking at the code.

core/server.go
```diff
@@ -647,6 +650,7 @@ func (m *RpcServer) GetWorkflowTemplates(cxt context.Context, req *pb.GetWorkflo
 					Repo: string(repo),
 					Revision: string(revision),
 					Template: template.Name,
+					Description: varSpecMap["description"].DefaultValue,
```
teo, line 653: I'm not sure this will work. The description is not a variable and should have no associated `VarSpec`. A `VarSpec` is a structure that declares a user-facing variable, but a description is a read-only property of a workflow template (and by extension, of an environment).

Have you tested this?

teo (changes requested):

core/workflow/iteratorrole.go
```diff
@@ -247,6 +247,13 @@ func (i *iteratorRole) GetName() string {
 	return i.template.GetName()
 }
 
+func (i *iteratorRole) GetDescription() string {
+	if i == nil || i.template == nil {
+		return ""
+	}
+	return i.template.GetDescription()
+}
+
```
teo, line 256: We don't need this part any more. You can remove the description field from "role" classes.

teo (changes requested):

teo (approved):

PR #332, opened 2022-03-15
[OCTRL-210] Report StartOfTransition/EndOfTransition/TransitionDuration timestamps in gRPC API

claireguyot:


core/server.go
```diff
@@ -368,6 +373,9 @@ func (m *RpcServer) ControlEnvironment(cxt context.Context, req *pb.ControlEnvir
 		Id: env.Id().String(),
 		State: env.CurrentState(),
 		CurrentRunNumber: env.GetCurrentRunNumber(),
+		StartOfTransition: sot.Format(time.RFC3339Nano),
+		EndOfTransition: eot.Format(time.RFC3339Nano),
+		TransitionDuration: strconv.FormatFloat(td.Seconds(), 'f', 6, 64),
```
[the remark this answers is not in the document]
graduta: I think we should look at keeping a standards across the gRPC files;

Right now I can see that AliECS returns dates as string in the following format: `2022-03-15T21:02:20+01:00`. I took the example from `GetEnvironment` and `GetEnvironments` methods. ( `string createdWhen = 2;` ). Maybe this was chosen as this format so that it goes well with `coconut`?

I would be a fan of starting using normal timestamps from now on and we bring the already existing string ones to this format. This would be what you suggested @teo `int64`

teo (approved):

core/protos/o2control.proto
```diff
@@ -176,7 +176,7 @@ message GetEnvironmentsReply {
 }
 message EnvironmentInfo {
     string id = 1;
-    string createdWhen = 2;
+    int64 createdWhen = 2;
```
teo, line 179: @graduta please note the backwards-incompatible change.

graduta: @teo  thank you for the notification. What is the expected delivery date for this?

teo: @graduta this week's release. The only mandatory change for you is that `createdWhen` is now an int64 with nanoseconds since UNIX epoch. 3 additional fields were added but you're not forced to adapt to them immediately.

core/protos/o2control.proto
```diff
@@ -233,6 +233,9 @@ message ControlEnvironmentReply {
     string id = 1;
     string state = 2;
     uint32 currentRunNumber = 3;
+    int64 StartOfTransition = 4;
+    int64 EndOfTransition = 5;
+    int64 TransitionDuration = 6;
```
teo, line 238: @claireguyot could you please make these lowercase for consistency? You should then regenerate the code, though I expect the output to be identical.

PR #338, opened 2022-03-24
[core] Ensure all timestamps sent by AliECS are in milliseconds

teo:


teo:
@claireguyot FYI, I had to make this change urgently in order to tag.

teo:
@graduta All timestamps now in msec, you should be able to parse these

teo:
CI fails because of old Go, 1.17 is now required as a minimum. Force merging.

PR #341, opened 2022-03-29
[core] Kafka plugin to publish FSM state transitions

knopers8:


core/integration/kafka/protos/kafka.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
```
knopers8, line 1: @awegrzyn let me know if the format & data suits you

core/integration/kafka/plugin.go
```diff
@@ -0,0 +1,271 @@
...
+	return ""
+}
+
+func (p *Plugin) FSMTransitionTopic(state string) string {
+	return "FSM_" + state
+}
+
+func (p *Plugin) ActiveRunsListTopic() string {
+	return "ACTIVE_RUNS"
+}
```
knopers8, line 96: @awegrzyn here are the topic names, let me know if you prefer something different.

awegrzyn: For me it's fine, if you want to go for "industry" standard you can take a look at: https://cnr.sh/essays/how-paint-bike-shed-kafka-topic-naming-conventions

core/integration/kafka/plugin.go
```diff
@@ -0,0 +1,271 @@
...
+	err = json.Unmarshal(bytes, &detectorsSlice)
+	if err != nil {
+		log.WithError(err).
+			Error("error processing the detectors list")
+		return
+	}
+	return detectorsSlice, nil
+}
+
+func (p *Plugin) NewEnvStateObject(varStack map[string]string, state string) *kafkapb.EnvInfo {
```
teo, line 126: TBD I guess, but in principle if `state` is the actual current AliECS environment state machine state, there is probably a nicer way to get it than ask the user to call with the correct string param. See `call.go:208` and further: we preload some special values in the plugin call context, for each call, at the time of call instantiation. Right now I see no impediment to adding e.g. `__call_current_fsm_state` or similar.

core/integration/kafka/plugin.go
```diff
@@ -0,0 +1,271 @@
...
+	} else {
+		log.WithField("partition", envId).Debugf("Kafka message delivered message to topic %s [%d] at offset %v\n",
+			*m.TopicPartition.Topic, m.TopicPartition.Partition, m.TopicPartition.Offset)
+	}
+}
+
+func (p *Plugin) CreateUpdateCallback(varStack map[string]string, state string) func() string {
+	return func() (out string) {
+		// Retrieve and update the env info
+		timestamp := uint64(time.Now().UnixNano() / 1000000)
```
teo, line 218: We now depend on Go 1.17 or later, so it'd be best to use `.UnixMilli()` for readability

teo (changes requested):

core/integration/kafka/protos/kafka.proto
awegrzyn: Looks good to me, just 1 comment, I think it might be handy if information whether this is `SOR`/`EOR`... could be also included in the message. Eg. when you subscribe to multiple topics it's way easier to read it from field rather than looking from which topic it comes from.

core/integration/kafka/plugin.go
awegrzyn: Actually it'd be the best if topics were configurable instead of hardcoded. On the other side there's advantage, if new state is introduced there's no need to change the code. If we keep it as it is, how do I find out full list of topics?

awegrzyn:
Could you generate binary before merging it so I could take a look how it works?

core/integration/kafka/protos/kafka.proto
knopers8: I am afraid we don't have access to the last transition from within "enter_state" callback, thus I do not know myself if the FSM reaches e.g. CONFIGURED from CONFIGURE or STOP.
Perhaps @teo could confirm. Maybe we could store the last transition in `varStack`?

core/integration/kafka/plugin.go
knopers8: As far as I understand, topic names are not something you should change often, because one cannot rename an existing one.
I am not sure if this is what you are looking for, but this lists all the topics:
```
[root@piotr-kafka ~]# /opt/kafka/kafka_2.12-2.8.1/bin/kafka-topics.sh --bootstrap-server localhost:9092 --list
ACTIVE_RUNS
FSM_CONFIGURED
FSM_DEPLOYED
FSM_RUNNIN
FSM_RUNNING
__consumer_offsets
notification
```
If we follow a more "industrial" convention, they should be quite easy to filter out.

awegrzyn: Okay indeed, this should be good enough

core/integration/kafka/protos/kafka.proto
awegrzyn: I was referring to `state` variable that you also use to select the topic.

knopers8: There is `state` in `EnvInfo`.

core/integration/kafka/plugin.go
knopers8: Regarding the topic names, perhaps something like this instead?
- fsm transitions: `streaming.env_state.<state>` (notice that `state` is stored in upper case by aliecs, but i can convert it to lower case to adhere to the rest)
- active runs: `streaming.env_list.running`

core/integration/kafka/plugin.go
knopers8: This came out to be less trivial than expected, because the code in call.go:208 is not aware of the environment state and it is not clear to me how to propagate it. One can obtain a parent role state, which is not the same (and is always UNKNOWN in this case). Thus, I took a different approach to extract the state from the `trigger` string, less clean, but still not requiring to provide `state` upfront.

I would propose to merge it as it is and once you are back, we can improve it.

core/integration/kafka/plugin.go
knopers8: @awegrzyn :arrow_up:

awegrzyn: No worries, it can be upper case, my only comment is that `streaming` does bring any useful information, so I'd go for `aliecs.env.<state>` and `aliecs.env.running` or even skipping `aliecs.`.

knopers8: Agreed about `streaming`, but I believe we should differentiate between `env_state` and `env_list`. The first provides updates about new states, the latter contains lists of environments in given state (now only `running`).
If we used `env`, we would have `aliecs.env.RUNNING` for new running state notification and `aliecs.env.running` for the list of environments in `RUNNING`

awegrzyn: ok, indeed

awegrzyn:
@knopers8 Under `aliecs.env_list.RUNNING` topics I can find list of running environments, it'd be very useful to get list of created environments in addition to that.

awegrzyn:
@knopers8 I restarted AliECS core during ongoing run, but unfortunately I have not received empty active runs list after the restart. This is crucial, otherwise after core restart Big Screen may still display ongoing runs.

core/integration/kafka/plugin.go
```diff
@@ -0,0 +1,311 @@
...
+	timestamp := uint64(time.Now().UnixMilli())
+	activeRunsList := &kafkapb.ActiveRunsList{
+		ActiveRuns: p.GetRunningEnvList(),
+		Timestamp:  timestamp,
+	}
+	arlData, err := proto.Marshal(activeRunsList)
+	if err != nil {
+		log.Error("Could not marshall an active runs list: ", err)
+	}
+	p.ProduceMessage(arlData, p.ActiveRunsListTopic(), "")
```
teo, line 121: I suspect some of these member functions could be made private (uncapitalized initial)

core/integration/kafka/plugin.go
```diff
@@ -0,0 +1,311 @@
...
+			*m.TopicPartition.Topic, m.TopicPartition.Partition, m.TopicPartition.Offset)
+	}
+}
+
+func (p *Plugin) CreateUpdateCallback(varStack map[string]string) func() string {
+	return func() (out string) {
+		// Retrieve and update the env info
+		timestamp := uint64(time.Now().UnixMilli())
+		envInfo := p.NewEnvStateObject(varStack)
+		p.UpdateRunningEnvList(envInfo)
```
teo, line 260: As hinted before these should all be unexported.

core/integration/kafka/plugin.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+	varStack := call.VarStack
+
+	stack = make(map[string]interface{})
+	stack["Standby"] = p.CreateUpdateCallback(varStack)
+	stack["Deployed"] = p.CreateUpdateCallback(varStack)
+	stack["Configured"] = p.CreateUpdateCallback(varStack)
+	stack["Running"] = p.CreateUpdateCallback(varStack)
+	stack["Error"] = p.CreateUpdateCallback(varStack)
+	stack["Destroy"] = p.CreateUpdateCallback(varStack)
```
teo, line 304: I see that the callback builder only depends on the `varStack`. In this context, what's the impediment to just having a single key-value in the `stack` of functions, e.g. `stack["PublishUpdate"] = p.CreateUpdateCallback(varStack)` ?

core/server.go
```diff
@@ -391,7 +391,11 @@ func (m *RpcServer) ControlEnvironment(cxt context.Context, req *pb.ControlEnvir
 	td := eot.Sub(sot)
 
 	if err != nil {
-		env.Sm.SetState("ERROR")
+		err := env.TryTransition(environment.NewGoErrorTransition(m.state.taskman))
```
teo, line 394: Since you now use it, perhaps it would be prudent to amend the transition table in `environment.go` so that `GO_ERROR` is allowed from all states.

core/server.go
```diff
@@ -391,7 +391,11 @@ func (m *RpcServer) ControlEnvironment(cxt context.Context, req *pb.ControlEnvir
...
+		if err != nil {
+			log.WithField("partition", env.Id()).Warn("we could not transition gently to ERROR, thus forcing it.")
```
teo, line 396: Please rephrase and wrap: `Warnf("could not complete requested GO_ERROR transition, forcing move to ERROR: %w", err)`

teo (changes requested):

core/integration/kafka/plugin.go
knopers8: None, i actually realised that as well, but forgot after :facepalm:. Will do

core/server.go
knopers8: done, but used `err.Error()`, because `%w` may be used only with `Errorf`

teo (approved):

PR #351, opened 2022-04-28
[core] Port kafka plugin to kafka-go package

teo:


knopers8 (approved):
All seems good, thank you for the quick development

PR #360, opened 2022-05-05
[core][OCTRL-574] Set odc_topology_fullname as a runtime variable

claireguyot:


core/integration/odc/plugin.go
```diff
@@ -637,6 +639,33 @@ func (p *Plugin) ObjectStack(varStack map[string]string) (stack map[string]inter
...
+			if !ok {
+				log.WithField("partition", envId).
+					WithField("call", "GenerateEPNWorkflowScript").
+					Error("cannot acquire ODC topology variable")
+				return
+			}
+			odc_topology_fullname = "(xml, " + odc_topology + ")"
+
+		default:
+			odc_topology_fullname = ""
```
teo, line 661: The whole `default` clause here is unnecessary, you already initialise it to `""`.

teo (changes requested):

teo (approved):

PR #366, opened 2022-05-13
[core] Add auto-stop functionality for environments

kostorr:


core/environment/environment.go
```diff
@@ -834,3 +847,44 @@ func (env *Environment) GetVarsAsString() string {
 	}
 	return sortMapToString(varStack)
 }
+
+// return true if the auto stop transition has been scheduled, false otherwise
+// if true, the expected stop time is also returned
+func (env *Environment) scheduleAutoStopTransition() (bool, time.Time) {
+	autoStopEnabled, ok := env.UserVars.Get("auto_stop_enabled")
```
teo, line 854: This way you're only querying the `UserVars` (exclusively user input) of the global context, so you risk missing some of the variable sources. Instead I recommend using `env.GetKV("" /*empty path*/, "auto_stop_enabled")`.

core/environment/environment.go
```diff
@@ -834,3 +847,44 @@ func (env *Environment) GetVarsAsString() string {
...
+	if autoStopEnabled == "true" && ok {
+		var autoStopTimeout string
+		autoStopTimeout, ok = env.UserVars.Get("auto_stop_timeout")
```
teo, line 857: Same as above.

core/environment/environment.go
```diff
@@ -834,3 +847,44 @@ func (env *Environment) GetVarsAsString() string {
...
+				env.autoStopTimer = time.NewTimer(autoStopDuration)
+				go func() {
+					select {
+					case <-env.autoStopTimer.C:
+						log.WithField("partition", env.id).
+							WithField("run", env.currentRunNumber).
+							Infof("Executing scheduled auto stop transition following expiration of %s", autoStopDuration)
+						err = env.TryTransition(NewStopActivityTransition(ManagerInstance().taskman))
+						if err != nil {
+							fmt.Println(err)
```
teo, line 872: If you have something to print, please use `log.Debug` or similar instead.

teo (changes requested):

teo (approved):

PR #369, opened 2022-05-16
[core][O2B-560] Adding required fields for Runs in Bookkeeping

claireguyot:


teo (dismissed):

vascobarroso:
@claireguyot @teo I guess this needs to be synchronized with changes in the bookkeeping API right ?

claireguyot:
@vascobarroso Indeed, however the priority is the LHC fills for Bookkeeping right now, so it has to wait for a bit before the necessary changes can be made.

core/integration/bookkeeping/api.go
```diff
@@ -66,7 +66,7 @@ func Instance() *BookkeepingWrapper {
 	return instance
 }
 
-func (bk *BookkeepingWrapper) CreateRun(activityId string, nDetectors int, nEpns int, nFlps int, runNumber int32, runType string, timeO2Start time.Time, timeTrgStart time.Time, dd_flp bool, dcs bool, epn bool, epnTopology string, detectors string) error {
+func (bk *BookkeepingWrapper) CreateRun(activityId string, nDetectors int, nEpns int, nFlps int, runNumber int32, runType string, dd_flp bool, dcs bool, epn bool, epnTopology string, detectors string) error {
```
kostorr, line 69: Pedantic comment: Is there a reason for snake case on `dd_flp`? If not I'd use camel case to keep it consistent

core/integration/bookkeeping/api.go
```diff
@@ -80,26 +80,23 @@ func (bk *BookkeepingWrapper) CreateRun(activityId string, nDetectors int, nEpns
 		runtypeAPI = sw.TECHNICAL_RunType
 	}
 
-	_, _, err := clientAPI.CreateRun(activityId, int32(nDetectors), int32(nEpns), int32(nFlps), runNumber, runtypeAPI, timeO2Start, timeTrgStart, dd_flp, dcs, epn, epnTopology, sw.Detectors(detectors))
+	_, _, err := clientAPI.CreateRun(activityId, int32(nDetectors), int32(nEpns), int32(nFlps), runNumber, runtypeAPI, dd_flp, dcs, epn, epnTopology, sw.Detectors(detectors))
```
kostorr, line 83: same for all occurences of `dd_flp`

core/integration/bookkeeping/plugin.go
```diff
@@ -179,8 +179,10 @@ func (p *Plugin) CallStack(data interface{}) (stack map[string]interface{}) {
 		epn_enabled, _ := strconv.ParseBool(env.GetKV("", "epn_enabled"))
 		odc_topology := env.GetKV("", "odc_topology")
 		detectors := strings.Join(env.GetActiveDetectors().StringList(), ",")
+		//odc_topology_fullname, _ := env.Workflow().GetVars().Get("odc_topology_fullname")
+		//lhc_period := env.GetKV("", "lhc_period")
 
-		err = p.bookkeepingClient.CreateRun(env.Id().String(), len(env.GetActiveDetectors()), 0, len(flps), int32(runNumber), env.GetRunType().String(), time.Now(), time.Now(), dd_enabled, dcs_enabled, epn_enabled, odc_topology, detectors)
+		err = p.bookkeepingClient.CreateRun(env.Id().String(), len(env.GetActiveDetectors()), 0, len(flps), int32(runNumber), env.GetRunType().String(), dd_enabled, dcs_enabled, epn_enabled, odc_topology, detectors)
```
kostorr, line 185: Same for variables here, just a convention coming from somewhere else?

core/integration/bookkeeping/plugin.go
```diff
@@ -234,9 +236,52 @@ func (p *Plugin) CallStack(data interface{}) (stack map[string]interface{}) {
 		}
 		return
 	}
-	updateRunFunc := func(runNumber64 int64, state string) (out string) {
+	updateRunFunc := func(runNumber64 int64, state string, timeO2Start time.Time, timeO2End time.Time, timeTrgStart time.Time, timeTrgEnd time.Time) (out string) {
 		callFailedStr := "Bookkeeping UpdateRun call failed"
-		err := p.bookkeepingClient.UpdateRun(int32(runNumber64), state, time.Now(), time.Now())
+		//odc_topology_fullname, _ := env.Workflow().GetVars().Get("odc_topology_fullname")
+		trg_global_run_enabled, err := strconv.ParseBool(env.GetKV("", "trg_global_run_enabled"))
+		if err != nil {
```
kostorr, line 243: In this group of parses if a single parse fails the update function exits; wouldn't we want to update even if some field is missing?

core/integration/bookkeeping/plugin.go
```diff
@@ -284,10 +330,11 @@ func (p *Plugin) CallStack(data interface{}) (stack map[string]interface{}) {
 		}
 
 		envState := env.CurrentState()
+
 		if envState != "RUNNING" {
-			return updateRunFunc(runNumber64, "bad")
+			return updateRunFunc(runNumber64, "test", time.Now(), time.Time{}, time.Now(), time.Time{})
```
kostorr, line 335: What is the significance of "test" in this context?

core/integration/bookkeeping/api.go
claireguyot: No, there isn't. You're right, I'll change this!

core/integration/bookkeeping/plugin.go
claireguyot: I was wondering about this, but I guess you're right, I will change this to not exit on parse error.

core/integration/bookkeeping/plugin.go
claireguyot: It is the default for the run quality, as per what George told me

kostorr (approved):

PR #370, opened 2022-05-18
[cleanup-binary] Synchronize after kills

kostorr:


kostorr:
@teo @vascobarroso this blocks only if necessary and until the processes are no longer present. Handles zombie and background processes gracefully as well.

teo (approved):

PR #373, opened 2022-06-01
[core] Better run error handling in Bookkeeping plugin

claireguyot:


core/integration/bookkeeping/plugin.go
```diff
@@ -120,6 +120,13 @@ func (p *Plugin) CallStack(data interface{}) (stack map[string]interface{}) {
 		log.Error("cannot acquire environment ID")
 		return
 	}
+	trigger, ok := varStack["__call_trigger"]
+	if !ok {
+		log.WithField("call", call).
+			WithField("partition", envId).
+			Error("cannot acquire trigger from varStack")
+		return
```
teo, line 128: In principle, every time you wish to return with an error condition from an integration plugin function, you should set `__call_error` as a minimum on the calling context `VarStack`, and ideally also `__call_error_reason`, like so:
```
			call.VarStack["__call_error_reason"] = err.Error()
			call.VarStack["__call_error"] = callFailedStr
```

It's not ideal or idiomatic, but still the best way to relay error values between the template and caller contexts.

If you don't do this, the environment state machine doesn't know that your call failed (the return is just a return with string type), and keeps going even when we'd want the env to fail and notify the user or InfoLogger.

teo (changes requested):

core/integration/bookkeeping/plugin.go
claireguyot: Ah yes, I understand. So, do we want the environment to fail when we can't retrieve the `trigger` for the `varStack` in the Bookkeeping plugin?

teo: No (or not yet anyway), and in any case we don't get to decide this here: generally an integration plugin should have no power to terminate an environment, it can only say "all good" (no `__call_error`) or "this or that error happened" (`__call_error` plus `__call_error_reason`). In the latter case, the only thing that's interrupted is the current integration plugin call.

Then it's up to the `critical` trait defined in the workflow template to decide whether the failure of this call also triggers an environment failure.

teo (changes requested):

core/integration/bookkeeping/plugin.go
```diff
@@ -117,7 +118,22 @@ func (p *Plugin) CallStack(data interface{}) (stack map[string]interface{}) {
 	varStack := call.VarStack
 	envId, ok := varStack["environment_id"]
 	if !ok {
-		log.Error("cannot acquire environment ID")
+		err := errors.New("cannot acquire environment ID")
```
kostorr, line 121: Similar comment to the other PR: should these parse failures return an error? I cannot judge. Otherwise looks good

core/integration/bookkeeping/plugin.go
```diff
@@ -117,30 +118,60 @@ func (p *Plugin) CallStack(data interface{}) (stack map[string]interface{}) {
...
+			Error(err)
+
+		call.VarStack["__call_error_reason"] = err.Error()
+		call.VarStack["__call_error"] = "Bookkeeping plugin Call Stack failed"
 		return
 	}
 	var err error
 	parsedEnvId, err := uid.FromString(envId)
 	if err != nil {
+		if strings.Contains(trigger, "DESTROY") || strings.Contains(trigger, "GO_ERROR") {
```
kostorr, line 142: Shouldn't these (multiple occurences) also populate the `__call_error` and `__call_error_reason`?

kostorr (changes requested):

core/integration/bookkeeping/plugin.go
claireguyot: No, these should not, it's one of the points of the PR

kostorr (approved):

PR #377, opened 2022-06-14
[OCTRL-667][core] Write down EOR time also on RUNNING->ERROR

knopers8:


knopers8:
Verified it works with https://github.com/AliceO2Group/ControlWorkflows/pull/374

teo (approved):

PR #378, opened 2022-06-14
[auto-stop] Transition into ERROR if stop fails

kostorr:


core/environment/environment.go
```diff
@@ -880,7 +880,8 @@ func (env *Environment) scheduleAutoStopTransition() (scheduled bool, expected t
 					if err != nil {
 						log.WithField("partition", env.id).
 							WithField("run", env.currentRunNumber).
-							Errorf("Scheduled auto stop transition failed: %s", err.Error())
+							Errorf("Scheduled auto stop transition failed: %s\n Transitioning into ERROR", err.Error())
```
teo, line 883: Log output should have no `\n`, as that's "end of message" for InfoLogger.

teo (changes requested):

teo (approved):

PR #379, opened 2022-06-17
[config] Empty repos ssh key by default

kostorr:


kostorr:
Set by ansible on consul only at P2 for the mirror.

PR #393, opened 2022-07-12
[OCTRL-679] O2 times and Trg times should be the correct ones for BK runs

claireguyot:


core/integration/trg/plugin.go
```diff
@@ -702,6 +705,9 @@ func (p *Plugin) CallStack(data interface{}) (stack map[string]interface{}) {
 			WithField("runNumber", runNumber64).
 			Debug("TRG RunStop success")
 
+		trgEndTime := strconv.FormatInt(time.Now().UnixMilli(), 10)
+		call.GetParentRole().(callable.ParentRole).SetRuntimeVar("trg_end_time_ms", trgEndTime)
```
teo, line 709: Same as above

core/integration/bookkeeping/plugin.go
```diff
@@ -344,8 +344,7 @@ func (p *Plugin) CallStack(data interface{}) (stack map[string]interface{}) {
 			return
 		} else {
 			var updatedRun string
-			emptyTime := time.Time{}
-			if timeO2Start == emptyTime {
+			if function, ok := varStack["__call_func"]; ok && function == "UpdateRunStop" {
```
teo, line 347: I believe `function == "UpdateRunStop"` risks failing (I think the content of `__call_func` might include "`bookkeeping.`" as well) you should use `strings.Contains(function, "UpdateRunStop")` instead.

teo (changes requested):

PR #397, opened 2022-07-15
[core] TRG plugin returns the TRG Stop timestamp on RunStop and Cleanup

claireguyot:


core/integration/trg/plugin.go
```diff
@@ -602,6 +602,7 @@ func (p *Plugin) CallStack(data interface{}) (stack map[string]interface{}) {
 		parentRole, ok := call.GetParentRole().(callable.ParentRole)
 		if ok {
 			parentRole.SetGlobalRuntimeVar("trg_start_time_ms", trgStartTime)
+			parentRole.DeleteGlobalRuntimeVar("trg_end_time_ms")
```
teo, line 605: This is all very good, my only proposal is to call this same function near the top of `RunLoad` to clear both start and end timestamps, before the `"not a TRG Global Run, continuing with TRG Run Start"` return so the vars are cleared in every case. This way you execute any future steps of this run with cleared TRG timestamp information.

teo (changes requested):

teo (approved):

PR #410, opened 2022-08-10
[OCTRL-683] [executor] Allow the OCClib-controlled tasks to exit on their own

knopers8:
Two problems are solved here:
- OccServer::EventStream thread would not exit when reaching DONE,
  thus making OCClib-controlled tasks never shut their server down
- executor would send SIGTERM as soon as a task reported reaching
  DONE, thus not allowing it to exit on its own.


executor/executable/controllabletask.go
```diff
@@ -417,13 +418,13 @@ func (t *ControllableTask) Launch() error {
 			}
 			for {
 				if t.rpc == nil {
-					log.WithField("partition", t.knownEnvironmentId.String()).
+					log.WithField("partition", t.knownEnvironmentId.String()).WithField("taskId", deo.TaskId.String()).
```
teo, line 421: Can you please wrap each `WithField` on its own line? I know it's a silly nitpick but it makes these log statements much more readable.
Otherwise we're good to go :rocket:

teo (changes requested):

executor/executable/controllabletask.go
knopers8: no worries, fair point!

teo (approved):

PR #411, opened 2022-08-12
WIP: [OCTRL-649] Allowing non-critical tasks

claireguyot:


core/environment/transition_deploy.go
```diff
@@ -169,6 +169,9 @@ func (t DeployTransition) do(env *Environment) (err error) {
 
 		return*/
 
+	////////////////
+	// CHECK HERE //
+	////////////////
```
teo, line 174: No task exist yet here, probably not a useful spot

core/task/state.go
```diff
@@ -64,6 +64,9 @@ func StateFromString(s string) State {
 	return UNKNOWN
 }
 
+////////////////
+// CHECK HERE //
+////////////////
```
teo, line 69: This is just the arithmetic of state changes

core/workflow/aggregator.go
```diff
@@ -48,20 +48,23 @@ func (r *aggregator) copy() copyable {
...
 	*iteratorRole
 	*aggregatorRole
 	*taskRole
 	*callRole
 	*includeRole
 }
 
+////////////////
+// CHECK HERE //
+////////////////
```
teo, line 67: Critical trait already unmarshaled

core/workflow/safestate.go
```diff
@@ -25,23 +25,34 @@
...
 	state task.State
 }
 
-func aggregateState(roles []Role) (state task.State) {
+func aggregateState(roles []Role, s task.State) (state task.State) {
 	if len(roles) == 0 {
 		state = task.INVARIANT
 		return
 	}
-	state = roles[0].GetState()
```
teo, line 42: In this case, knowing `s` as the state of an unknown task that last changed is useless, because you anyway have to go through all children in order to decide a new aggregate state.

The difficulty here is that unlike before when `state` could a priori be initialized from the first child role (whichever it may be), this is no longer true because `state` can only be initialized from the first **critical** child role, as states of non-critical roles must be ignored.

2 ways to go about it:

1:
* start going through `roles` and find the first critical one
* initialize `state` to it
* iterate on `range roles[idx+1:]` where `idx` is the index of the first critical child role.

2:
* make `state` a `*task.State` instead of `task.State`, init to `nil`
* start iterating and check for role type and criticality as before
* if state is `nil`, initialize, otherwise do `MIXED`/`ERROR`/`X()` as before. 

If there are no critical tasks at all, return `INVARIANT`.

teo (changes requested):

teo (approved):

PR #415, opened 2022-08-19
Refactor of component configuration access layer

teo:
Extended and improved configuration path resolution, it can now happen in the workflow template context at any time, with or without a full KV query
Support for ANY/any as fallback paths for run type and role
Extended per-role template processing sequence, it now operates on 5 stages (up from 4), and it processes the `enabled` condition before doing anything else in order to improve correctness, reliability and performance
Added support for string + <nil> and <nil> + string operations in the template system for reliability
A task's special varStack is now built as early as possible so task templates can use task_hostname and similar runtime values
Expanded template system API and split into config/strings/dpl/util units (old calls are deprecated but still available for backwards compatibility with pre-v0.50.0 AliECS), changes include:
* GetConfig -> config.Get
* new functions config.Resolve/ResolvePath/ResolveWithTimestamp/ResolvePathWithTimestamp
* DetectorForHost/DetectorsForHosts etc. -> inventory.DetectorForHost/DetectorsForHosts
* GenerateDplSubworkflow/GenerateDplSubworkflowFromUri -> dpl.Generate/GenerateFromUri
* GetRuntimeConfig -> runtime.Get, plus corresponding new function runtime.Set to push a value to the component's runtime context in Apricot/Consul
* Atoi/Itoa/ToUpper etc. -> strings.Atoi/Itoa/ToUpper
* ToJson/FromJson -> json.Marshal/Unmarshal
* NewID -> uid.New
* PrefixedOverride/Dump -> util.PrefixedOverride/Dump plus new function util.Nullable to safely wrap a value that might be <nil>


apricot/protos/apricot.pb.go
```diff
@@ -48,40 +48,69 @@ const (
 // of the legacy proto package is being used.
 const _ = proto.ProtoPackageIsVersion4
 
+// NOTE: make sure the enum values include and match those in RunType in dcs.pb.go and runtype.go
+// NOTE: this run type list is replicated in AliceO2 repo in GRPECSObject.h. Inform Ruben when the list is updated.
```
knopers8, line 52: remember to let Ruben know ;)

configuration/template/stack.go
```diff
@@ -55,224 +49,112 @@ type ToPtreeFunc func(string, string) string
...
-			log.Warn(varStack)
-			log.Warn(payload)
-			return payload
+func MakeConfigAccessObject(confSvc ConfigurationService, varStack map[string]string) map[string]interface{} {
+	obj := make(map[string]interface{})
+	obj["config"] = map[string]interface{}{
+		"Get": func(path string) string {
+			return getConfig(confSvc, varStack, path)
 		},
-		"GetConfig": func(path string) string {
```
knopers8, line 81: Please keep the old names for some time, as these are used in automatically generated templates. Can we have some transition period?
See https://github.com/AliceO2Group/AliceO2/blob/dev/Framework/Core/src/O2ControlHelpers.cxx#L220

knopers8 (commented):
I can't give it a proper review before the release, but I agree with the main message. See my two comments though

configuration/template/stack.go
teo: Yes this is not a problem, the old names will be kept indefinitely but new functions will only be added in the new namespaces.

PR #418, opened 2022-09-02
[OCTRL-701] Configuration files can be loaded from consul

claireguyot:


coconut/control/control.go
```diff
@@ -278,15 +279,30 @@ func CreateEnvironment(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.C
...
-		if err != nil {
-			err = errors.New("cannot read configuration payload file")
-			return err
-		}
-		err = json.Unmarshal(configPayloadDoc, &payloadData)
-		if err != nil {
-			err = errors.New("cannot unmarshal configuration payload")
-			return err
+		if strings.HasSuffix(strings.ToLower(configPayload), ".json") {
+			configPayloadDoc, err := os.ReadFile(configPayload)
```
teo, line 283: You are overwriting this error in the next line. Please make sure all `err` values are checked.

coconut/control/control.go
```diff
@@ -278,15 +279,30 @@ func CreateEnvironment(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.C
...
+			err = json.Unmarshal(configPayloadDoc, &payloadData)
+			if err != nil {
+				err = errors.New("cannot unmarshal local configuration payload")
+				return err
+			}
+		} else {
+			configPayloadSplit := strings.Split(configPayload, "/")
+			if len(configPayloadSplit) == 0 {
```
teo, line 291: I believe this will only ever be true on a `nil` string? Even a non-`nil` empty string with `Split` will yield `[""]`, therefore `len == 1`. What are you aiming to check for here? The validity of the Consul subpath? It's hard to do that before actually performing the Consul query, and the `Split` for `/` won't tell you whether it's a valid Apricot runtime path or not.

coconut/control/control.go
```diff
@@ -278,15 +279,30 @@ func CreateEnvironment(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.C
...
+				err = errors.New("configuration payload file is not valid")
+				return err
+			}
+			configPayloadUri := configPayloadSplit[len(configPayloadSplit)-1]
```
teo, line 295: This too might be problematic, because within their runtime subpath, components can do anything including create subtrees. In other words, both `foo` and `bar/quux` are valid inputs to `-c`, because both `o2/runtime/COG-v1/foo` and `o2/runtime/COG-v1/bar/quux` are valid paths (and should work assuming payloads exist at those keys). On the other hand, we a priori exclude Consul full paths (i.e. `consul://...`) because Apricot `GetRuntimeEntry` doesn't allow us to read from anywhere.

coconut/control/control.go
```diff
@@ -278,15 +279,30 @@ func CreateEnvironment(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.C
...
+			configPayloadDoc, err := apricot.Instance().GetRuntimeEntry("COG-v1", configPayloadUri)
```
teo, line 296: Minor nitpick: please put `"COG-v1"` in a `const HLCONFIG_RUNTIME_COMPONENT_PREFIX` on top of the file, or similar.

coconut/control/control.go
```diff
@@ -278,15 +279,30 @@ func CreateEnvironment(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.C
...
+			if err != nil {
+				err = errors.New("cannot retrieve file from consul under COG-v1 component")
```
teo, line 298: ... and then please also use it here, as well as fixing the error being clobbered here too (see next comment for details).

coconut/control/control.go
```diff
@@ -278,15 +279,30 @@ func CreateEnvironment(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.C
...
+				return err
+			}
+			err = json.Unmarshal([]byte(configPayloadDoc), &payloadData)
+			if err != nil {
+				err = errors.New("cannot unmarshal remote configuration payload")
```
teo, line 303: Clobbering the error, we risk undetected bugs this way.

Please see https://go.dev/blog/go1.13-errors it's a short but important article describing the changes to the `errors` package in Go 1.13, and specifically the concept of wrapping, which allows you to build your error objects like matryoshka dolls and never lose lower-level error information.

In this specific case, your `errors.New` should be replaced with `fmt.Errorf` and the `%w` operator.

coconut/control/control.go
```diff
@@ -278,15 +279,30 @@ func CreateEnvironment(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.C
...
-		}
-		err = json.Unmarshal(configPayloadDoc, &payloadData)
-		if err != nil {
-			err = errors.New("cannot unmarshal configuration payload")
-			return err
+		if strings.HasSuffix(strings.ToLower(configPayload), ".json") {
+			configPayloadDoc, err := os.ReadFile(configPayload)
+			err = json.Unmarshal(configPayloadDoc, &payloadData)
+			if err != nil {
+				err = errors.New("cannot unmarshal local configuration payload")
```
teo, line 286: Clobbering the error here too, please see comment on wrapping at the bottom.

teo (changes requested):

coconut/control/control.go
```diff
@@ -262,56 +265,62 @@ func GetEnvironments(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.Com
 func CreateEnvironment(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.Command, args []string, o io.Writer) (err error) {
 	configPayload, err := cmd.Flags().GetString("configuration")
```
teo, line 266: Initially I thought this was minor and went ahead but I had to go back up to address it because it actually confuses me quite a bit when reading the code: you're calling it `configPayload` but it's actually a `configPayloadKey` or `configPayloadPath` or `configPayloadIdentifier` or `configPayloadEntryName` (pick whichever you prefer). When I write "payload", I'm thinking "content", "value", not "identifier": we use the latter to then fetch the former.

coconut/control/control.go
```diff
@@ -262,56 +265,62 @@ func GetEnvironments(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.Com
...
+			if err != nil {
+				return fmt.Errorf("cannot unmarshal local configuration payload: %w", err)
+			}
+		} else {
+			if strings.HasPrefix(configPayload, HLCONFIG_PATH_PREFIX+HLCONFIG_COMPONENT_PREFIX) {
+				configPayload = strings.TrimPrefix(configPayload, HLCONFIG_PATH_PREFIX+HLCONFIG_COMPONENT_PREFIX)
+			}
+			configPayloadDoc, err := apricot.Instance().GetRuntimeEntry(HLCONFIG_COMPONENT_PREFIX, configPayload)
+			if err != nil {
+				return fmt.Errorf("cannot retrieve file from "+HLCONFIG_PATH_PREFIX+HLCONFIG_COMPONENT_PREFIX+": %w", err)
```
teo, line 296: I'd also include the actual key we're failing to retrieve (i.e. the contents of `configPayload` that should really be called `configPayloadKey` or somesuch). Also, if it's on disk we call it a "file", if it's from Consul we call it a "payload", "entry" or "document".

Generally speaking about data resources:
* if it's on disk it's a file and it has a name;
* if it's in memory (in Consul, in a buffer, object storage, etc.) then it's a payload or document, and it has a key or path.

coconut/control/control.go
```diff
@@ -262,56 +265,62 @@ func GetEnvironments(cxt context.Context, rpc *coconut.RpcClient, cmd *cobra.Com
...
+				return errors.New("no workflow template provided in config file")
 			}
 		}
 	} else {
 		if cmd.Flags().Changed("workflow-template") && len(userWfPath) > 0 {
 			wfPath = userWfPath
 		} else {
-			err = errors.New("empty workflow template provided")
-			return err
+			return errors.New("empty workflow template provided")
```
teo, line 323: This is really minor, but generally we should avoid having identical log messages (or identical error messages) arising in different spots, because this makes it hard to understand *where* something failed.
Here we are in the "no -c, empty -w" scenario, and the message is fine, but could you please modify the instance of this message on L308 so that it makes it clear that an empty -w was passed as override to a configuration payload, and that this is empty.

teo (changes requested):

teo (approved):

PR #419, opened 2022-09-02
[OCTRL-710][core] do not include newline char in JIT config queries

knopers8:
If the config file address was put at the end of JIT command, as below:
```
o2-qc --config json://path/to/payload
```
then the regex would include a newline character in the matched string,
which would not pass `inputFullRegex` when building a query. However,
we would try to retrieve the configuration anyway, because there was no
error check after invoking `NewQuery`.

### DIFF
--- a/configuration/componentcfg/query.go
+++ b/configuration/componentcfg/query.go
@@ -65,6 +65,7 @@ func NewEntriesQuery(path string) (p *EntriesQuery, err error) {
 		RunType:   apricotpb.RunType_NULL,
 		RoleName:  "",
 	}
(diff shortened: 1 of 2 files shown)


teo:
:exploding_head:

teo:
Good catch @knopers8 . However I'm still a bit concerned, are we risking creating a `Query` object with one of its members still having a tailing newline? Perhaps you could sanitise the whole input through `strings.TrimSpace` before `Split`ting it and processing it further.

knopers8:
Yes, we still risk this in principle, adding trimming.

PR #424, opened 2022-10-03
[executor] log signals when killing tasks

knopers8:
This way we can be sure that SIGINT, SIGTERM, SIGKILL was sent by AliECS, not by a different entity.


knopers8:
closing in favour of https://github.com/AliceO2Group/Control/pull/425

PR #428, opened 2022-10-11
DCS proto file - added AGD in Detector enum

dvoscek:


core/integration/dcs/protos/dcs.proto
```diff
@@ -204,6 +204,7 @@ enum Detector {
     ZDC           = 15;
     LHC           = 16;
     DCS           = 17;
+    AGD           = 18;
```
teo, line 207: We already have `TST` as a special value defined [here](https://github.com/AliceO2Group/Control/blob/677f939282512846e59bdaf3bd98f82f5ac306e9/common/system/systemid.go#L64). Its purpose seems to be the ECS equivalent of what `AGD` is being proposed for: a placeholder detector code, that triggers code paths with "generic logic" to simulate a detector without hardware.

Would you be willing to consider using `TST` for this purpose on the DCS side? This would allow us to avoid the proliferation of special-purpose system IDs if the use cases fit.

teo (changes requested):

core/integration/dcs/protos/dcs.proto
dodovoscek: I get your point, but AGD on our side is not simulation without hardware, it is real hardware that serves testing purposes and is embedded in DCS architecture, that's why we would like to have AGD defined as a real detector in the communication protocol.

teo: Thanks @dvoscek, this clarifies a bit. Then I guess the question is for the FLP team: **what do we do with `TST`**? 

Options:
* We keep both `AGD` and `TST` in ECS, with `AGD` alone involving DCS simulation hardware, and `TST` treated as a "do nothing" input by DCS.
* We remove `TST` and only have `AGD` in AliECS, for use as a dummy detector both with DCS simulation, and *without* DCS at all (open question: would the removal of `TST` cause trouble in Bookkeeping, O² tasks or other system components?).
* We don't add `AGD` at all, anywhere in ECS, and only keep `TST`, but if `TST` is included and the DCS integration client is enabled, the client sends `AGD` to DCS instead of `TST`. In this case `AGD` would remain exclusively a DCS concept and not an entity in the ECS or elsewhere.

@vascobarroso any thoughts/preferences on this?

vascobarroso: @teo @dvoscek from these options I think I prefer the last bullet point.

dodovoscek: @teo @vascobarroso but then we will have 2 versions of this proto file, one with TST(on your side) and one with AGD(on our side), I don't see this as a good idea for the future

teo: @dvoscek sorry, I'm afraid I haven't explained this well: the protofile will stay one and the same, and exclusively with `AGD` as you proposed. It is on me as ECS maintainer to ensure that if `TST` is included and DCS enabled, all calls made to DCS include `AGD` instead of `TST`.

dvoscek: @teo thanks for the clarification, so it was my misunderstanding, all clear now

PR #430, opened 2022-10-14
[core] Conversion between ECS and DCS detector codes in DCS client

teo:


teo:
@dvoscek this is the follow up to your previous PR, where we make sure that an ECS selection of `TST` results in `AGD` being included in the DCS SOR and EOR.

PR #436, opened 2022-11-04
[executor] do not try to recover a task in error, ask to exit instead

knopers8:
Could not test every possible scenario, but it seems to be fine.

### DIFF
--- a/executor/executable/controllabletask.go
+++ b/executor/executable/controllabletask.go
@@ -614,8 +614,8 @@ func (t *ControllableTask) Kill() error {
 				evt = "RESET"
 				destination = "STANDBY"
 			case "ERROR":
-				evt = "RECOVER"
-				destination = "STANDBY"
+				evt = "EXIT"
+				destination = "DONE"
 			case "STANDBY":
 				evt = "EXIT"
 				destination = "DONE"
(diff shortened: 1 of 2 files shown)


executor/executable/controllabletask.go
```diff
@@ -614,8 +614,8 @@ func (t *ControllableTask) Kill() error {
 				evt = "RESET"
 				destination = "STANDBY"
 			case "ERROR":
-				evt = "RECOVER"
-				destination = "STANDBY"
+				evt = "EXIT"
+				destination = "DONE"
```
teo, line 618: Shouldn't we still also allow the `RECOVER` transition anyway? I get it doesn't exist right now as an actual operation but IIRC the FSM accounts for it.

teo (changes requested):

executor/executable/controllabletask.go
knopers8: yes, you're right

knopers8: No, but wait. We still do allow for it in `OccServer.cxx`, this has not changed. What has changed is that we EXIT instead of RECOVER when in ERROR in `Kill()`. Do you mean that we should try to RECOVER once before we EXIT? I do not see the point of that if we need to kill the tasks anyway.

teo: @knopers8 no, you're right, your change is correct. Thanks, merging.

teo (approved):

PR #439, opened 2022-11-10
[OCTRL-718][apricot] Trim whitespaces in vars from Consul, and return warning

claireguyot:


apricot/local/serviceutil.go
```diff
@@ -84,7 +85,13 @@ func (s *Service) getStringMap(path string) map[string]string {
 			if v.Type() != cfgbackend.IT_Value {
 				continue
 			}
-			theMap[k] = v.Value()
+			if v.Value() == "\n" || k == "" || k == "\n" {
```
teo, line 88: We're getting there but not quite yet.

Firstly, here if the value is `"\n"`, you're just printing a warning but you aren't returning the value. The values should still be returned even if a warning is needed.

Secondly, you're also checking the key and not just the value, which was not required, and in fact is incorrect. We cannot just decide to return the value of a trimmed key or something like that (basically decreeing that a key with or without whitespace is the same), because then the assumptions behind what a *key* means are negated (e.g. uniqueness).

Thirdly, you are checking if the whole value is `"\n"`, which is not what we want, we discussed this. What you need is to check *whether the value has leading or trailing whitespace characters*, and if yes, warn and still use the value. One way to do this is with `string.HasPrefix/HasSuffix` calls, another, perhaps simpler, is to compare the string with the output of `strings.TrimSpace`. We also discussed returning the trimmed value or the verbatim value depending on a Viper setting to be added to Apricot as a final step, this is only possible if you indeed return some value after the check.

teo (changes requested):

apricot/local/serviceutil.go
```diff
@@ -84,7 +87,19 @@ func (s *Service) getStringMap(path string) map[string]string {
...
 				continue
 			}
-			theMap[k] = v.Value()
+			if v.Value() != strings.TrimSpace(v.Value()) {
+				valueToStore := v.Value()
+				log.WithError(errors.New("leading or trailing space in value")).
+					WithField("path", path).
+					WithField("key", k).
+					Warning("Consul var has leading/trailing space in its value")
+				if viper.GetBool("trimSpaceInVarsFromConsulKV") {
```
teo, line 96: I propose you extract this `GetBool` to right outside the `for`, to avoid the multiplication of queries to the viper singleton.

apricot/local/serviceutil.go
```diff
@@ -84,7 +87,19 @@ func (s *Service) getStringMap(path string) map[string]string {
 			if v.Type() != cfgbackend.IT_Value {
 				continue
 			}
-			theMap[k] = v.Value()
+			if v.Value() != strings.TrimSpace(v.Value()) {
```
teo, line 90: Better:
`if trimmed = strings.TrimSpace(v.Value()); v.Value != trimmed {`

This way you already have a handle to the trimmed string and you don't need to trim again on line 97.

apricot/local/serviceutil.go
```diff
@@ -84,7 +87,19 @@ func (s *Service) getStringMap(path string) map[string]string {
...
+				valueToStore := v.Value()
+				log.WithError(errors.New("leading or trailing space in value")).
```
teo, line 92: This is the error text, it is a fact about the payload, and we keep it.

apricot/local/serviceutil.go
```diff
@@ -84,7 +87,19 @@ func (s *Service) getStringMap(path string) map[string]string {
...
+					WithField("path", path).
+					WithField("key", k).
+					Warning("Consul var has leading/trailing space in its value")
```
teo, line 95: This part (the IL message) depends on the viper setting.
If false: `Warning("found unsanitized string value in configuration backend, returning anyway")`
If true: `Warning("found unsanitized string value in configuration backend, returning whitespace-trimmed string")`

InfoLogger will then append the `error` message as well.

teo (changes requested):

teo (approved):

PR #442, opened 2022-11-17
[OCTRL-649][core] Correcting aggregate state in case of MIXED/ERROR

claireguyot:


core/workflow/safestate.go
```diff
@@ -51,13 +52,20 @@ func aggregateState(roles []Role) (state task.State) {
...
+			rolesToCheck = append(rolesToCheck, c)
 		}
-		state = state.X(c.GetState())
+	}
+	if len(rolesToCheck) > 0 {
+		checkedState := state
+		for _, c := range rolesToCheck {
+			checkedState = checkedState.X(c.GetState())
+		}
+		state = checkedState
```
teo, line 68: Not sure this is the best way. You're setting aside some roles that happened to result in `state` being either `MIXED` or `ERROR`, and then recalculating `state` based on these roles. However there's no guarantee that any of these are either `MIXED` or `ERROR` , they just arrived after the first `MIXED` or `ERROR` had already been joined into `state`.

This should all be doable with a single loop, and specifically you *need* to establish precedence between `ERROR` and `MIXED` (`ERROR` trumps `MIXED` because `ERROR` means "at least one critical child is `ERROR`", whereas `MIXED` means "at least one critical child is in a different state than another child", a weaker condition altogether).

Have you considered establishing a bool, e.g. `hasChildrenInError`, to be asserted if necessary during the loop? Then in the loop you only ever need to check for `MIXED`, and after the loop `if MIXED && hasChildrenInError` tells you whether a response of `ERROR` state is warranted.

teo (changes requested):

core/workflow/safestate.go
claireguyot: @teo is it sufficient like this then? It doesn't seem to cause any issue with my limited testing scenario.

teo (approved):

PR #444, opened 2022-11-23
[doc][core] Documentation for non-critical tasks

claireguyot:


docs/handbook/configuration.md
```diff
@@ -1 +1,17 @@
-# Workflow and Task Configuration
\ No newline at end of file
+# Workflow and Task Configuration
+
+## Non-critical tasks
+
+Any task in a workflow can be declared as non-critical. A non-critical task is a task that doesn't trigger a global environment ERROR in case of failure. The state of a non-critical task doesn't affect the environment state in any way.
+
+To declare a task as non-critical, a line has to be added in the corresponding workflow template. Under the task role, in the `task` section (usually after the `load` statement), the line to add is `critical: false`, like in the following example:
```
teo, line 7: "...in the corresponding task role block within a workflow template file. Specifically, in the `task` section of such a task role (usually ..."

docs/handbook/configuration.md
```diff
@@ -1 +1,17 @@
...
+
+```yaml
+roles:
+  - name: "non-critical-task"
+    vars:
+      non-critical-task-var: 'var-value'
+    task:
+      load: mytask
+      critical: false
+```
```
teo, line 17: Very good, please add at the end:

In the absence of an explicit `critical` trait for a given task role, the assumed default value is `critical: true`.

teo (changes requested):

PR #446, opened 2022-11-24
[executor] truncate task command in logs when it will be cut

knopers8:
This truncates the long DPL commands to leave the part after the last pipe, which is the most crucial to understand what is running. We attempt to print the full command in "launching task" log. Also, the PR removes "cmd" field in logs where it is empty anyway.

### DIFF
--- a/executor/executable/controllabletask.go
+++ b/executor/executable/controllabletask.go
@@ -28,6 +28,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"github.com/AliceO2Group/Control/executor/executorutil"
 	"io"
 	"io/ioutil"
 	"reflect"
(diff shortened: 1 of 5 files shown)


knopers8:
an example log after the change:
```
executor<->occplugin interface requesting transition command="source /etc/profile.d/o2.sh && o2-dpl-raw-proxy -b --session default --dataspec 'x:TST/RAWDATA;dd:FLP/DISTSUBTIMEFRAME/0' --readout-proxy '--channel-config "name=readout-proxy,type=pull,method=connect,address=ipc:///tmp/stf-builder-dpl-pipe-0,transport=shmem,rateLogging=10"' | o2-dpl-output-proxy --environment DPL_OUTPUT_PROXY_ORDERED=1 -b --session default --dataspec 'x:TST/RAWDATA;dd:FLP/DISTSUBTIMEFRAME/0' --dpl-output-proxy '--channel-config "name=downstream,type=p <CUT> | o2-dpl-raw-proxy" event="INIT TASK" id="2bvjBnVsGva" task="github.com/AliceO2Group/ControlWorkflows/tasks/jit-79ebe3ddb92a9ce845b31c5fb773cd5267381966-internal-dpl-clock@ccc5aa4d530f12261cf033ea8aa6ff9c06820346?2bvjBnVsGva"
```

teo (approved):

PR #447, opened 2022-11-25
[OCTRL-704][core] template helpers for assigning JIT workflows per groups of FLPs

knopers8:
Here is an example of how they could be used: https://github.com/AliceO2Group/ControlWorkflows/pull/533/files

I hope that `HostnameWithin` reduces 150 string comparisons and `and` operations into a few for each FLP. In principle, we could even compare raw strings with IDs without converting them, assuming they use leading zeros, but I am not sure I want to risk it.

### DIFF
--- a/configuration/template/dplutil.go
+++ b/configuration/template/dplutil.go
@@ -137,14 +137,22 @@ func generateDplSubworkflow(confSvc ConfigurationService, varStack map[string]st
 	return jitDplGenerate(confSvc, varStack, workflowRepo, "source /etc/profile.d/o2.sh &&"+dplCommand)
 }
 
-func generateDplSubworkflowFromUri(confSvc ConfigurationService, varStack map[string]string, workflowRepo repos.IRepo, dplCommandUri string) (jitWorkflowName string, err error) {
+func generateDplSubworkflowFromUri(confSvc ConfigurationService, varStack map[string]string, workflowRepo repos.IRepo, dplCommandUri string, fallbackToTemplate bool) (jitWorkflowName string, err error) {
 	if dplCommandUri == "none" {
 		return "", fmt.Errorf("dplCommand is 'none'")
 	}
 
 	dplCommand, err := workflowRepo.GetDplCommand(dplCommandUri)
 	if err != nil {
-		return "", fmt.Errorf("Failed to read DPL command from '%s': %w\n", dplCommandUri, err)
+		if fallbackToTemplate {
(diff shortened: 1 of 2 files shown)


configuration/template/stack.go
```diff
@@ -316,6 +319,32 @@ func MakeUtilFuncMap(varStack map[string]string) map[string]interface{} {
 			}
 			return *in
 		},
+		"HostnameWithin": func(hostname string, prefix string, idMinStr string, idMaxStr string) string {
```
teo, line 322: Just a suggestion: perhaps we could call it `SuffixWithin` or `SuffixInRange` or somesuch? I realize you use it for hostnames but there's no reason to assume the input string must be a hostname.

configuration/template/dplutil.go
```diff
@@ -137,14 +137,22 @@ func generateDplSubworkflow(confSvc ConfigurationService, varStack map[string]st
...
 	}
 
 	dplCommand, err := workflowRepo.GetDplCommand(dplCommandUri)
 	if err != nil {
-		return "", fmt.Errorf("Failed to read DPL command from '%s': %w\n", dplCommandUri, err)
+		if fallbackToTemplate {
+			// if a file in JIT is missing, it will try to fallback to a standard workflow template in 'workflows/'.
+			// effectively, this allows us to have an intermediate switch workflow to select different JIT commands
+			// for different nodes.
+			log.Debugf("JIT: There is no file 'jit/%s' with a DPL command, falling back the template at 'workflows/%s'", dplCommandUri, dplCommandUri)
```
teo, line 151: Should this be a `Warnf` perhaps? Do you expect this to happen regularly, or only exceptionally when a `jit` request failed and needs to be fixed?

teo (changes requested):

configuration/template/stack.go
knopers8: Agreed. Initially I went for `HostnameWithin` so it is easier for users to realize what it could be used for, but if we have anyway an example of TPC, it should be clear enough.

configuration/template/dplutil.go
knopers8: This will happen regularly, so at each deployment with TPC we will have either 145 warnings or debug messages :/ I am not very fond of this fallback solution, but I do not see how to do it better without much greater effort.

knopers8:
@teo do you have any other comments?

teo (approved):

PR #448, opened 2022-11-28
[core] LeafWalk should also walk in include roles

knopers8:
This fixes error messages with missing failed roles:
```
cannot create new environment: transition canceled with error: workflow deployment failed, aborting and cleaning up [failed roles: ]
```
Here it is with the fix:
```
cannot create new environment: transition canceled with error: workflow deployment failed, aborting and cleaning up [failed roles: readout-dataflow.host-piotr-ostack3.data-distribution-dpl.dpl.internal-dpl-injected-dummy-sink]
```

### DIFF
--- a/core/workflow/roleutils.go
+++ b/core/workflow/roleutils.go
@@ -74,6 +74,11 @@ func Walk(root Role, do func(role Role)) {
 		for _, child := range typed.Roles {
 			LeafWalk(child, do)
 		}
+	case *includeRole:
+		do(typed)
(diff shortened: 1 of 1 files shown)


knopers8:
FYI, this is how we get the list of failed roles: https://github.com/AliceO2Group/Control/blob/master/core/environment/transition_deploy.go#L259

teo (approved):
Good catch, thanks @knopers8

PR #450, opened 2022-12-02
[OCTRL-709][core] pass flp id list to ccdb grp object

knopers8:
this works, but i am waiting on a decision in OCTRL-709 if we should accept only production FLPs or any FLPs. If the first, then the FLP prefix could be perhaps retrieved from some central value in consul or a header with constants at least...


knopers8:
ready for review

teo (approved):

PR #483, opened 2023-05-31
Compile with FairMQ 1.5.x

ktf:


ktf:
@teo ca. we have a new tag for this?

PR #488, opened 2023-06-26
[O2-2818][core] Disable autoBind in FairMQ

knopers8:
This makes FairMQ produce an error when failing to bind a port instead of selecting a random one in the range 22000-23000, which is useless in our case.

Probably not related to OCTRL-773, but should fix O2-2818, where we obliviously run a QC proxy which could not bind a port due to a leftover task on the same machine.


knopers8:
@teo do I have to change it anywhere else?

teo (approved):

PR #512, opened 2023-11-24
[core] Make sure end of trigger is before end of run

knopers8:


core/integration/ccdb/plugin.go
```diff
@@ -455,6 +455,21 @@ func (p *Plugin) CallStack(data interface{}) (stack map[string]interface{}) {
...
+				WithField("partition", envId).
+				WithField("level", infologger.IL_Ops).
+				Error("End of trigger time will be missing in the CCDB GRPECS. Please create a log entry in BK and tag PDP")
+		}
+		if trgEndTimeMsUint > o2EndTimeMsUint {
+			log.WithField("call", "RunStop").
+				WithField("partition", envId).
+				WithField("level", infologger.IL_Ops).
+				Errorf("End of trigger time will be later than end of run time (%d > %d) in the CCDB GRPECS. Please create a log entry in BK and tag PDP", trgEndTimeMsUint, o2EndTimeMsUint)
+		}
```
knopers8, line 472: @chiarazampolli @shahor02 what do you think of these messages?

I have to test the other fixes yet, so the PR is not ready as a whole.

PR #517, opened 2024-02-09
[apricot] implementation of GetRuntimeEntries rpc call

justonedev1:


apricot/local/service.go
```diff
@@ -477,6 +477,21 @@ func (s *Service) SetRuntimeEntry(component string, key string, value string) er
 	}
 }
 
+func (s *Service) GetRuntimeEntries(component string) ([]string, error) {
```
teo, line 480: I advise starting with `s.logMethod()` like the other handlers, just so the log output behaviour is consistent.

apricot/local/service.go
```diff
@@ -477,6 +477,21 @@ func (s *Service) SetRuntimeEntry(component string, key string, value string) er
...
+	if keys, err := s.ListRuntimeEntries(component); err != nil {
```
teo, line 481: I believe this should be the `err == nil` case

apricot/local/service.go
```diff
@@ -477,6 +477,21 @@ func (s *Service) SetRuntimeEntry(component string, key string, value string) er
...
+		entries := make([]string, 0)
+		for _, key := range keys {
+			if entry, err := s.GetRuntimeEntry(component, key); err != nil {
```
teo, line 484: This `err` is unhandled.

teo (changes requested):

apricot/local/service.go
```diff
@@ -477,6 +477,23 @@ func (s *Service) SetRuntimeEntry(component string, key string, value string) er
 	}
 }
 
+func (s *Service) GetRuntimeEntries(component string) (map[string]string, error) {
+	s.logMethod()
+
+	if keys, err := s.ListRuntimeEntries(component); err == nil {
+		entries := make(map[string]string)
+		for _, key := range keys {
+			if entry, err := s.GetRuntimeEntry(component, key); err == nil {
```
teo, line 486: This err is still unhandled.

In this kind of situation, elsewhere in AliECS we use the `go-multierror` package. Take a look for example at https://github.com/AliceO2Group/Control/blob/master/core/workflow/aggregatorrole.go#L171

teo (changes requested):

apricot/local/service.go
justonedev1: How exactly should it be handled? I thought that not adding entry that caused the error is enough. But according to multierror package you would prefer to return error from the function even if only one of X entries returns error and other are ok?

teo: Yes, that's the idea: if any number of entries were not returned because of an error, the errors should be packaged into a single error and returned. We still return the values we got, but we also inform the caller that something is amiss.

justonedev1: something like this?

apricot/local/service.go
```diff
@@ -481,13 +482,16 @@ func (s *Service) GetRuntimeEntries(component string) (map[string]string, error)
 	s.logMethod()
 
 	if keys, err := s.ListRuntimeEntries(component); err == nil {
+		var keysErrors *multierror.Error
 		entries := make(map[string]string)
 		for _, key := range keys {
 			if entry, err := s.GetRuntimeEntry(component, key); err == nil {
 				entries[key] = entry
+			} else {
+				keysErrors = multierror.Append(keysErrors)
```
teo, line 491: You're not appending anything into your keysErrors here, otherwise the idea is right.

teo (changes requested):

teo (approved):

PR #521, opened 2024-03-01
[core] Ensure the CTP readout FLP is included in BK call if enabled

teo:


core/integration/bookkeeping/plugin.go
```diff
@@ -431,6 +431,24 @@ func (p *Plugin) CallStack(data interface{}) (stack map[string]interface{}) {
...
+
+		// The enabled status of CTP Readout and its FLP is governed by the variable ctp_readout_enabled, so it requires
+		// special treatment to be included in the list of FLPs
+		ctpReadoutEnabled, err := strconv.ParseBool(env.GetKV("", "ctp_readout_enabled"))
+		if err != nil {
+			log.WithError(err).
+				WithField("run", runNumber64).
+				WithField("partition", envId).
+				WithField("call", "StartOfRun").
+				Warning("cannot parse CTP readout enabled")
```
knopers8, line 443:
```suggestion
				Warning("cannot parse ctp_readout_enabled")
```
For clarity, I would use the exact variable name, so users are not mislead that they are using a wrong variable name.

knopers8 (approved):

teo:
@graduta @martinboulais FYI

PR #524, opened 2024-03-05
[core] Make task manager eventStream thread safe

teo:


core/environment/eventStream.go
```diff
@@ -40,40 +39,45 @@ type Subscription interface {
...
+	defer e.mu.Unlock()
 	close(e.stream)
 	e.stream = nil
 }
 
 type eventSub struct {
-    feed         *eventStream
-    once         sync.Once
-    err          chan error
+	feed *eventStream
```
justonedev1, line 71: is there a reason for the change of format?

core/environment/environment.go
```diff
@@ -987,7 +987,13 @@ func (env *Environment) scheduleAutoStopTransition() (scheduled bool, expected t
...
 							WithField("run", env.currentRunNumber).
 							Errorf("Scheduled auto stop transition failed: %s, Transitioning into ERROR", err.Error())
-						_ = env.TryTransition(NewGoErrorTransition(ManagerInstance().taskman))
+						err = env.TryTransition(NewGoErrorTransition(ManagerInstance().taskman))
+						if err != nil {
+							log.WithField("partition", env.id).
+								WithField("run", env.currentRunNumber).
+								Errorf("Forced transition to ERROR failed: %s", err.Error())
+							env.setState("ERROR")
+						}
```
justonedev1, line 996: does this part of code do something for the thread safety?

justonedev1 (commented):
This seems ok if eventStream  is the only part that needs to be thread safe. However, I have few nitpicks, which I did not understand how those got there.

core/environment/environment.go
teo: No, this bit ensures we go to ERROR even if the GO_ERROR transition fails for any reason.

justonedev1 (approved):

PR #525, opened 2024-03-06
[core] Ensure auto-transitioning environments always end

teo:


justonedev1:
I would probably name the function differently then `goErrorKillDestroy` (something like `sendErrorAndCleanup`), but it the code seems ok for me :)

justonedev1 (approved):

PR #526, opened 2024-03-11
[core] hash processed config payloads from apricot during JIT

knopers8:
This commit makes the JIT template generation support both consul-json and apricot backends of the O2 Configuration library. Also, instead of the LastIndex, we use the processed configuration payload in the hash sum which is used to detect any relevant changes in the DPL workflows.

### DIFF
--- a/configuration/template/dplutil.go
+++ b/configuration/template/dplutil.go
@@ -32,7 +32,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
-	"strconv"
 	"strings"
 	texttemplate "text/template"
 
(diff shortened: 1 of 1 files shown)


knopers8:
Tested on an OpenStack setup with FLP suite:
- backwards-compatibility with existing workflows using consul-json backend
- correct detection of a switch between consul-json and apricot backend
- correct detection of an update in the requested payload
- correct detection of an update of another file included in the requested payload

JIT fails if the URI includes a REST query, such as `--config apricot://{{ apricot_endpoint }}/components/qc/ANY/any/stfb_to_daqtask-{{ it }}?process=true`. This will be reported as a bug, but since the `process` seems to be `true` by default, I do not consider this PR as blocked by the bug. It seems we need an update here: https://github.com/AliceO2Group/Control/blob/981cc042d8ba174afae901f6048f34207208e45b/configuration/componentcfg/query.go#L43

teo:
@knopers8 https://its.cern.ch/jira/browse/OCTRL-862

teo (approved):

PR #532, opened 2024-03-14
[core] support query parameters in apricot URIs in JIT

knopers8:


knopers8:
For '&' delimiters in query params, we need https://github.com/AliceO2Group/AliceO2/pull/12873

teo (approved):

PR #535, opened 2024-03-21
[core] test task.State, demonstrate a possible bug

knopers8:


core/task/state_test.go
```diff
@@ -0,0 +1,88 @@
...
+		})
+
+		When("one of the states is ERROR", func() {
+			It("returns ERROR", func() {
+				Expect(task.ERROR.X(task.UNKNOWN)).To(Equal(task.ERROR))
+				Expect(task.ERROR.X(task.RUNNING)).To(Equal(task.ERROR))
+				Expect(task.ERROR.X(task.INVARIANT)).To(Equal(task.ERROR))
+
+				Expect(task.UNKNOWN.X(task.ERROR)).To(Equal(task.ERROR)) // FIXME: this fails
+				Expect(task.RUNNING.X(task.ERROR)).To(Equal(task.ERROR)) // FIXME: this fails
```
knopers8, line 53: @teo I assume that `X()` stands for multiplication operation. Shouldn't it be commutative, so that ERROR x RUNNING returns the same as RUNNING x ERROR?

teo: It should!

teo (approved):

core/task/state_test.go
knopers8: cool, i'll bring a fix soon

teo (approved):

PR #536, opened 2024-03-21
[core] add safestate unit test, demonstrate two bugs

knopers8:


core/workflow/safestate_test.go
```diff
@@ -0,0 +1,158 @@
...
+				subroles[1] = &taskRole{
+					roleBase: roleBase{state: SafeState{state: task.CONFIGURED}},
+					Traits:   task.Traits{Trigger: "", Await: "", Timeout: "", Critical: true},
+				}
+				ar := &aggregatorRole{
+					roleBase:   roleBase{state: SafeState{state: task.CONFIGURED}},
+					aggregator: aggregator{subroles},
+				}
+				ar.state.merge(task.ERROR, ar)
+				Expect(ar.GetState()).To(Equal(task.CONFIGURED)) // fixme: this fails
```
knopers8, line 130: This seems to be related to https://its.cern.ch/jira/browse/OCTRL-853

core/workflow/safestate_test.go
```diff
@@ -0,0 +1,158 @@
...
+				subroles[1] = &taskRole{
+					roleBase: roleBase{state: SafeState{state: task.DONE}},
+					Traits:   task.Traits{Trigger: "", Await: "", Timeout: "", Critical: true},
+				}
+				ar := &aggregatorRole{
+					roleBase:   roleBase{state: SafeState{state: task.RUNNING}},
+					aggregator: aggregator{subroles},
+				}
+				ar.state.merge(task.DONE, ar)
+				Expect(ar.GetState()).To(Equal(task.MIXED))
```
knopers8, line 149: This behaves as expected now, but according to the requirements stated in https://its.cern.ch/jira/browse/OCTRL-846, the aggregated state should go to ERROR as long as the expected state is RUNNING. I think we will have to introduce a new concept of an expected state for the tasks...

core/workflow/safestate_test.go
```diff
@@ -0,0 +1,158 @@
...
+					roleBase: roleBase{state: SafeState{state: task.MIXED}},
+					Traits:   task.Traits{Trigger: "", Await: "", Timeout: "", Critical: true},
+				}
+				ar := &aggregatorRole{
+					roleBase:   roleBase{state: SafeState{state: task.ERROR}},
+					aggregator: aggregator{subroles},
+				}
+				ar.state.merge(task.MIXED, ar)
+				Expect(ar.GetState()).To(Equal(task.ERROR)) // FIXME: this fails
+			})
```
knopers8, line 62: @teo may i get your confirmation that this test case should not fail?

core/workflow/safestate_test.go
teo: @justonedev1 ↑

core/workflow/safestate_test.go
teo: Correct. `ERROR`×`ANYTHING` is still `ERROR`.

teo (approved):

PR #540, opened 2024-03-25
[occ] switched to cmake functions officially provided by libproto

justonedev1:
I switched from custom cmake functions created by @teo to cmake function provided by libprotobuf.
I provided `cmake` object library that contains all of the generated files (.proto, .pb.*, .grpc.pb.*), so it is easy to link against while keeping correct build and generate order.


justonedev1:
I am quite sure, that this will fix the race-condition issue when generating rpms, I would like to test it, but I don't know how. Can you give me a hint how to test it on build machines before merging? @teo @knopers8

teo:
> I am quite sure, that this will fix the race-condition issue when generating rpms, I would like to test it, but I don't know how. Can you give me a hint how to test it on build machines before merging? @teo @knopers8

@justonedev1 I don't think you can, but @vascobarroso would know

occ/CMakeLists.txt
```diff
@@ -430,9 +341,7 @@ add_library(${OCCPLUGIN} SHARED
 
 target_link_libraries(${OCCPLUGIN} PUBLIC
     FairMQ::FairMQ
-    gRPC::grpc++
-    gRPC::grpc++_reflection
-    protobuf::libprotobuf
+    Occ-proto
```
teo, line 344: Is an additional linkage to a new .so the only way?

justonedev1: It does not create new .so. Occ-proto is new object library, which as exactly designed for these purposes so the build structure is not polluted by additional shared/static libraries

vascobarroso:
> > I am quite sure, that this will fix the race-condition issue when generating rpms, I would like to test it, but I don't know how. Can you give me a hint how to test it on build machines before merging? @teo @knopers8
> 
> @justonedev1 I don't think you can, but @vascobarroso would know

@Barthelemy can you answer the question above ?

justonedev1:
@knopers8 proposed to the this https://gitlab.cern.ch/AliceO2Group/system-configuration/-/blob/dev/ansible/docs/developers/flp-suite/procedures/flp-suite-release.md?ref_type=heads

knopers8:
To be precise, I proposed a test release on Michal's fork, which then could be used in points 2-4 of the procedure linked by Michal.

Barthelemy:
That would work. 

We can also try by hand on the release machine but it is never easy to know if it really fixes a race condition.

justonedev1:
Can I run the pipeline multiple times, even if successful? I know that I can not be really 100% sure if it fixes race condition, but trying it multiple times is the best way I can think off.

teo (approved):

PR #542, opened 2024-03-27
Never propagate non-critical task state to parent, add tests

knopers8:
More details in the description of specific commits.


justonedev1 (approved):
I really like this and huge thumbs up for the tests!

teo (approved):

PR #543, opened 2024-03-27
[core] remove obsolete workaround to make ERROR state always win

knopers8:
As far as I understand, this was an attempt at fixing the cases when ERROR state is overriden by MIXED, which was fixed in https://github.com/AliceO2Group/Control/pull/535 and https://github.com/AliceO2Group/Control/pull/536 directly in the state arithmetics code. I don't think this is needed anymore and happy tests seem to confirm it.

### DIFF
--- a/core/workflow/safestate.go
+++ b/core/workflow/safestate.go
@@ -39,7 +39,6 @@ func aggregateState(roles []Role) (state task.State) {
 	if len(roles) == 0 {
 		return
 	}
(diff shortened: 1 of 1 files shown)


knopers8:
It's fine to postpone for some time until we are sure it does not break some corner case, but unless we find something, it would be good to remove it, so we do not wonder what it was for in 5 years.

teo (approved):

PR #544, opened 2024-03-28
[core] Emit role events to aliecs.role topic

teo:


knopers8 (commented):
The code makes sense, I am only wondering what is going to be the performance hit in the production system. Have you made any considerations?

teo:
@knopers8 the event emission is asynchronous and lock-free, if there's any contention it'll be for those `t.status.get()` calls. Ultimately we need to try in prod, and revert if it turns out slow.

knopers8 (approved):

PR #545, opened 2024-03-28
OCTRL-851 [core] do not attempt to transition inactive tasks

knopers8:
This does not solve the case when the task is in ERROR, but alive, but it is already an improvement.

### DIFF
--- a/core/environment/transition_configure.go
+++ b/core/environment/transition_configure.go
@@ -26,6 +26,7 @@ package environment
 
 import (
 	"errors"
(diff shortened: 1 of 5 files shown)


justonedev1 (approved):
If you are saying that this is an improvement to existing state, than let's go ahead and try it. Pr looks fine

knopers8:
@teo I took the liberty to proceed with this one in order to have a pre-release for staging (v1.8.1), where I would like it to run for some time before it goes into prod.

PR #546, opened 2024-04-04
added coverage step to makefile

justonedev1:


Makefile
```diff
@@ -153,6 +153,18 @@ test:
...
 
+coverage:
+	@echo -e "[Ginkgo] \033[1;33mgo test -coverprofile=gotestginkgo.out -v\033[0m $(TEST_DIRS)\033[0m"
+	@$(BUILD_FLAGS) go test -coverprofile=gotestginkgo.out -v $(TEST_DIRS) -ginkgo.show-node-events
+	@echo -e "[Ginkgo] \033[1;33mgo tool cover -html=gotestginkgo.out -o gotestginkgo.html\033[0m"
+	@$(BUILD_FLAGS) go tool cover -html=gotestginkgo.out -o gotestginkgo.html
+
+	@echo -e "\n[gotest] \033[1;33mgo test -v\033[0m $(GO_TEST_DIRS)\033[0m"
+	@$(BUILD_FLAGS) go test -coverprofile=gotest.out -v $(GO_TEST_DIRS)
+	@echo -e "[Ginkgo] \033[1;33mgo tool cover -html=gotest.out -o gotest.html\033[0m"
```
knopers8, line 164:
```suggestion
	@echo -e "[gotest] \033[1;33mgo tool cover -html=gotest.out -o gotest.html\033[0m"
```
I assume you want `gotest` label here.

knopers8 (changes requested):
Thanks a lot! I tried it on my laptop and it seems to work nicely. I did also notice that we are getting a lot of noise from failing to send kafka events, we should fix that separately.

I am wondering if for the sake of test coverage we should run all the test at once instead of splitting them into ginkgo and non-ginkgo outputs. This way we should be able to get one HTML report instead of two.

justonedev1:
You are right that it would be better to probably merge these tests into one. But I have to admit, that I was not looking into differences between these 2 as I just wanted to see how easy/hard it is to create coverage in golang. But I can look into it as a part of this PR and if possible just to merge these two tests into one

knopers8:
I quickly checked myself and it seems the easiest is to just run all the tests at once to get one `.out`. There is a go package which might be able to merge those, but it seems to be not maintained since 8 years...

justonedev1:
I did a small workaround after testing multiple approaches. The file format of coverage is just a text file with line describing which line was tested or not. So I just added these output files together into `merged.out` which I than convert to `merged.html` I kinda don't like this, but it seems simple enough that it works. 

I tried few packages that are quite old and `go tool covdata merge`, which said that it cannot merge created files...

knopers8:
Could you please comment why did you exclude the approach to run all the test at once? This seems to work well for me:
```
162         @echo -e "\n[gotest] \033[1;33mgo test -v\033[0m $(GO_TEST_DIRS) $(TEST_DIRS)\033[0m"
163         @$(BUILD_FLAGS) go test -coverprofile=gotest.out -v $(GO_TEST_DIRS) $(TEST_DIRS)
164         @echo -e "[Ginkgo] \033[1;33mgo tool cover -html=gotest.out -o gotest.html\033[0m"
165         @$(BUILD_FLAGS) go tool cover -html=gotest.out -o gotest.html
```

justonedev1:
Does this work? doesn't it just rewrite existing file?

justonedev1:
You are right, I was trying that as well, but I thought that ginkgo test flags are necessary for tests to run, but they just increase verbosity of outputs, which we don't really care for in coverage report

knopers8 (approved):
Thank you!

PR #547, opened 2024-04-10
[core] keeping created kafka writers in global object

justonedev1:
This is probably not the best way how to fix the issue of creating kafka writers every time we want to send the message. But I am using it as a draft to demonstrate the fix in the ticket. We will discuss with @knopers8 whether this is the best way to fix this.

### DIFF
--- a/core/the/eventwriter.go
+++ b/core/the/eventwriter.go
@@ -25,14 +25,47 @@
 package the
 
 import (
+	"sync"
+
 	"github.com/AliceO2Group/Control/common/event"
 	"github.com/AliceO2Group/Control/common/event/topic"
+	"github.com/AliceO2Group/Control/common/logger"
+	"github.com/sirupsen/logrus"
 )
 
+var (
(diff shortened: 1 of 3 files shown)


justonedev1:
I think that it is ready now

core/the/eventwriter.go
```diff
@@ -25,14 +25,47 @@
...
+func Clear() {
+	mu.Lock()
+	defer mu.Unlock()
+
+	log.Logf(logrus.InfoLevel, "Clearing %d kafka producers", len(writers))
+	for _, writer := range writers {
+		writer.Close()
+	}
+	clear(writers)
 }
```
knopers8, line 71: I would propose to either rename it to `ClearEventWriter` since it does only that. If we want it to clear all the singletons, we can refactor it later, but for now the call:
```
the.Clear()
```
is misleading.

Otherwise, i think we could move on with this PR.

knopers8 (approved):
thank you!

PR #548, opened 2024-04-11
[doc] update the release documentation

knopers8:


vascobarroso:
@knopers8 I think steps 10 and 11 can be removed.
Step 10. is obsolete (we don't use anymore jenkins to generate RPMs)  and it is covered in dedicated docs pages. 
Step 11. is no longer needed because we now use updatable RPMs.

knopers8:
ok, very good, i'll update also that

teo (approved):

PR #550, opened 2024-04-22
[core] Added missing mutex around eventStream in environment

justonedev1:
While searching for the cause of [OCTRL-882](https://its.cern.ch/jira/browse/OCTRL-882) I found this error with missing mutex around channel write. It is in the middle of all the symptoms (more in the ticket), so it even might be the cause of the issue, but I cannot be sure about that without more information when the aliecs gets stuck.

As I cannot be sure that this PR will fix mentioned bug, I propose to keep this PR not merged until we will be able to demonstrate the issue with somebody being able to `delve` into the aliecs status while it gets stuck and get stack trace of all of goroutines.


knopers8 (approved):
Since the last patch release contains a fix (golang bump) that should address the issue with the unresponsive core, I think we could target this fix for the next release.

PR #553, opened 2024-04-26
golang bump to 1.22

justonedev1:


knopers8 (approved):
for the record, this solves the issue with 824 billion tasks seen in `GetEnvironments` regardless of the architecture used to build the core.

PR #554, opened 2024-04-26
[common] OCTRL-805 allow to group entries in subfolders

knopers8:
This demonstrates the feature requested in OCTRL-805.

However, I am not sure if this does not collide with specifying the timestamp after an entry and `/`. I am not sure if this is feature is used by anyone, as I have never seen an example of that nor seen it documented. I will let @teo judge.

I tested it on an OS instance with some realistic QC configuration, it works.

If I have enough time before I leave on holidays and if this PR as in idea is approved, I can rework the regex queries to expect subentries explicitly, like the HTTP service does, for consistency.


knopers8:
This now includes the removal of the timestamp/versioning concept with some testing added, and is ready for review.

I might add some HTTP service tests as well if I have enough willpower. I did test it manually though.

teo (approved):

PR #560, opened 2024-05-03
[core] More granular mutex for task.Task fields

teo:


justonedev1 (commented):
I am not sure whether this is the way to do it properly. I would rather figure out why the setting of PID is hanging. 

Morover I don't think that this will for sure solve the issue as we are requesting id in serverutil.go `taskToShortTaskInfo` which is called on multiple places. So if the mutex will be locked like it seems to be in today's issue aliecs would still unresponsive in some calls...

teo:
> I am not sure whether this is the way to do it properly. I would rather figure out why the setting of PID is hanging.

Fair enough.

> Morover I don't think that this will for sure solve the issue as we are requesting id in serverutil.go `taskToShortTaskInfo` which is called on multiple places. So if the mutex will be locked like it seems to be in today's issue aliecs would still unresponsive in some calls...

Is this the first time that GetTaskPID popped up in the backtrace?

justonedev1:
no, it is in the ticket [OCTRL-882](https://its.cern.ch/jira/browse/OCTRL-882)

knopers8:
Do we still need this PR? Should it be closed?

PR #561, opened 2024-05-06
[core] added the ability to turn off kafka sending

justonedev1:


justonedev1:
I am not sure what is the purpose of Kafka plugin, so I put error inside. However, I don't know whether it is something we want @teo please let me know whether I should just change error to warning

teo (approved):

PR #564, opened 2024-05-22
Send SOSOR/EOSOR/SOEOR/EOEOR events to aliecs.run topic and improve timestamps lifetime handling

teo:
https://its.cern.ch/jira/browse/OCTRL-865
https://its.cern.ch/jira/browse/OCTRL-864
https://its.cern.ch/jira/browse/OCTRL-864


core/environment/environment.go
```diff
@@ -186,9 +186,22 @@ func newEnvironment(userVars map[string]string, newId uid.ID) (env *Environment,
 					env.workflow.GetVars().Set("run_number", rnString)
 					env.workflow.GetVars().Set("runNumber", rnString)
 
-					runStartTime := strconv.FormatInt(time.Now().UnixMilli(), 10)
-					env.workflow.SetRuntimeVar("run_start_time_ms", runStartTime)
-					env.workflow.SetRuntimeVar("run_end_time_ms", "") // we delete previous EOR
+					runStartTime := time.Now()
+					runStartTimeS := strconv.FormatInt(runStartTime.UnixMilli(), 10)
```
knopers8, line 190:
```suggestion
					runStartTimeStr := strconv.FormatInt(runStartTime.UnixMilli(), 10)
```
so it is not ambiguous with "S" for "Seconds"?

Same suggestion for similar occurrences elsewhere.

core/environment/environment.go
```diff
@@ -216,17 +229,43 @@ func newEnvironment(userVars map[string]string, newId uid.ID) (env *Environment,
...
 					}
 				} else if e.Event == "GO_ERROR" {
 					endTime, ok := env.workflow.GetUserVars().Get("run_end_time_ms")
 					if ok && endTime == "" {
-						runEndTime := strconv.FormatInt(time.Now().UnixMilli(), 10)
-						env.workflow.SetRuntimeVar("run_end_time_ms", runEndTime)
+						runEndTime := time.Now()
+						runEndTimeS := strconv.FormatInt(runEndTime.UnixMilli(), 10)
+						env.workflow.SetRuntimeVar("run_end_time_ms", runEndTimeS)
+						env.workflow.SetRuntimeVar("run_start_time_ms", "") // we delete previous SOSOR
```
knopers8, line 257: Why delete it here? It seems reasonable to me that tasks transitioning from RUNNING to CONFIGURED (due something else going to ERROR) might want to get all the known timestamps of the finishing run.
Perhaps you intended to delete EOEOR here?

core/environment/environment.go
```diff
@@ -439,7 +478,49 @@ func newEnvironment(userVars map[string]string, newId uid.ID) (env *Environment,
 							Infof("auto stop transition scheduled, expected execution at %s", expected)
 					}
 
+					runStartCompletionTime := time.Now()
+					runStartCompletionTimeS := strconv.FormatInt(runStartCompletionTime.UnixMilli(), 10)
+					env.workflow.SetRuntimeVar("run_start_completion_time_ms", runStartCompletionTimeS)
+					env.workflow.SetRuntimeVar("run_end_completion_time_ms", "") // we delete previous EOEOR
```
knopers8, line 484: Why deleting it at `before_START_ACTIVITY` is not enough?

core/environment/environment.go
```diff
@@ -439,7 +478,49 @@ func newEnvironment(userVars map[string]string, newId uid.ID) (env *Environment,
...
+						runEvent.TransitionStatus = pb.OpStatus_DONE_ERROR
+					}
+
+					the.EventWriterWithTopic(topic.Run).WriteEventWithTimestamp(runEvent, runStartCompletionTime)
+
 				} else if e.Event == "STOP_ACTIVITY" {
+					runEndCompletionTime := time.Now()
+					runEndCompletionTimeS := strconv.FormatInt(runEndCompletionTime.UnixMilli(), 10)
+					env.workflow.SetRuntimeVar("run_end_completion_time_ms", runEndCompletionTimeS)
+					env.workflow.SetRuntimeVar("run_start_completion_time_ms", "") // we delete previous EOSOR
```
knopers8, line 506: Why delete it already? I would not exclude that we could have a plugin which wants to retrieve the 4 timestamps only at `after_STOP_ACTIVITY`.

At least we could be coherent and delete either all the previous timestamps (SOSOR, EOSOR, SOEOR, EOEOR) or (perhaps better) keep them until the next `before_START_ACTIVITY`.

core/environment/manager.go
```diff
@@ -483,12 +483,43 @@ func (envs *Manager) TeardownEnvironment(environmentId uid.ID, force bool) error
...
+				Transition:       "TEARDOWN",
+				TransitionStatus: evpb.OpStatus_STARTED,
+				Vars:             nil,
+			}, runEndTime)
 		} else {
 			log.WithField("partition", environmentId.String()).
 				Debug("O2 End time already set before DESTROY")
 		}
+
+		endCompletionTime, ok := env.workflow.GetUserVars().Get("run_end_completion_time_ms")
```
knopers8, line 504: Do I understand correctly that any state changes in the controlled tasks will be perfomed only by the line?
```
taskmanMessage := task.NewEnvironmentMessage(taskop.ReleaseTasks, environmentId, tasksToRelease, nil)
```
i'm wondering what would be better of the two, to set an EOEOR which is in fact before we leave RUNNING, or to set EOEOR after the tasks have been transitioned to DONE... The later seems more correct to me, but I am not sure I even understand correctly what happens here.

knopers8 (commented):
Thanks! I have some questions, since I don't understand some decisions here, possibly something to change or at least clarify.

core/environment/manager.go
knopers8: OK, I see. Initially I read the code as the `DESTROY` hooks are called after having destroyed the tasks. It's only the message for releasing the tasks that is created before `DESTROY` hooks are called, but the message is only used after...

PR #565, opened 2024-05-22
OCTRL-888 [occ] If we see FairMQ's ERROR state, we should exit

knopers8:
FairMQ considers ERROR state as a final state and expects any plugins to exit in such case. This is of course in contradiction to our concept of the ERROR state, which we envisage to RECOVER from. This being said, since FairMQ tries to exit anyway in such cases and we do not plan to work on recovery soon, I think the best is to respect the current status quo. This fixes OCTRL-888, allowing the DPL tasks to exit upon seeing an exception and propagating it to FairMQ.


teo (approved):

teo:
@knopers8 good catch! I guess our `RECOVER`, if/when implemented, might at best redeploy the task.

knopers8:
I'm afraid so, I think we might be bound by what FairMQ will allow us to do.

PR #568, opened 2024-05-27
[core] Emit environment events for Teardown and AutoEnv

teo:
Includes:
* [OCTRL-869](https://its.cern.ch/jira/browse/OCTRL-869) Emit kafka aliecs.environment events for CreateAutoEnvironment
* Added missing event production for Teardown operation, in both AutoEnvironments and regular ones

@graduta backwards-compatible change in protofile. No intervention necessary, but allows you to eventually replace `CreateAutoEnvironment` with `CreateEnvironment`+`autoTransition=true`. The former is the old way with gRPC stream response, the latter behaves like the regular `CreateEnvironments` and expects you to consume from Kafka.


knopers8 (approved):
Thanks, I see no obvious issues.

I think the interface is a bit inconsistent since we require the GUI to use the `autoTransition` as a separate argument in the RPC call, but anyway it has to set `auto_stop_timeout` and `auto_stop_enabled` as user vars. I am wondering if all that should not be user vars for consistency... In any case, it's one GUI using it, not 100 users, so it does not change a lot.

PR #570, opened 2024-05-29
Ensure we push to the GUI the real time critical trait for taskRoles

teo:


core/serverutil.go
```diff
@@ -132,12 +132,12 @@ func taskToShortTaskInfo(t *task.Task, taskman *task.Manager) (sti *pb.ShortTask
 		},
 		Pid:           t.GetTaskPID(),
 		SandboxStdout: sandboxStdoutUri,
-		Critical:      t.GetTraits().Critical,
 	}
 	parentRole, ok := t.GetParentRole().(workflow.Role)
 	if ok && parentRole != nil {
 		sti.Status = parentRole.GetStatus().String()
 		sti.State = parentRole.GetState().String()
+		sti.Critical = parentRole.IsCritical()
```
knopers8, line 140: For my curiosity, why do you take these from the parent role?

knopers8 (approved):

core/serverutil.go
teo: Because that's where the critical trait is stored. The parent role of a `task.Task` (which is not a role) is a `taskRole`, accessed via the `workflow.Role` interface. Criticality is a property of the taskRole i.e. the leaf node of the workflow tree that points to the task in question. A task without a parentRole (insofar as it exists, and for the most part it shouldn't in current AliECS) has no criticality because criticality is a property of a task only in relation to its workflow.

In actuality the previous way to access the taskRole traits should've worked too (it also goes to the `taskRole` via `task.Task`, for which I'm about to commit a fix.

knopers8: thanks, clear.

knopers8 (approved):

PR #573, opened 2024-05-31
OCTRL-901 Document the order of actions performed during SOR and EOR

knopers8:
I think we are at the stage where things have settled down and there is a finite number of changes expected to such a document (namely: removing BKP and Kafka plugins calls, possibly moving CCDB plugin calls). One could even consider filling the hooks automatically, based on the information parsed from `readout-dataflow.yaml`, but I would leave that for later.

The reason for this document is also a request from users to know what are SOSOR, EOSOR, SOEOR, EOEOR timestamps exactly.

### DIFF
--- a/docs/handbook/operation_order.md
+++ b/docs/handbook/operation_order.md
@@ -21,18 +21,10 @@ This is the order of actions happening at a healthy start of run.
   - `kafka.PublishStartActivityUpdate()` at `-50`
   - `dcs.StartOfRun()`, `odc.Start()` (does not need to return now), `ccdb.RunStart()` at `0`
 
-### before_event
-
-- `before_event` hooks are executed
-
 ### leave_CONFIGURED
 
 - `leave_CONFIGURED` hooks are executed
   - `kafka.PublishLeaveStateUpdate()` at `0`
-
-### leave_state
-
-- `leave_state` hooks are executed
 
 ### Transition START_ACTIVITY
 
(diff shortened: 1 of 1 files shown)


docs/handbook/operation_order.md
```diff
@@ -0,0 +1,110 @@
...
+  - `trg.PrepareForRun()` at `-200`
+  - `trg.RunLoad()`, `bookkeeping.StartOfRun()` at `-100`
+  - `bookkeeping.RetrieveFillInfo()` at `-99`
+  - `kafka.PublishStartActivityUpdate()` at `-50`
+  - `dcs.StartOfRun()`, `odc.Start()` (does not need to return now), `ccdb.RunStart()` at `0`
+
+### before_event
+
+- `before_event` hooks are executed
+
```
teo, line 27: Thank you for starting this. I've double checked now and it looks like we do not support generic targets like `before_event` or `leave_state`, all the targets must be explicit like `before_START_ACTIVITY` and `leave_CONFIGURED`. The order reported in [configuration.md](configuration.md#State-machine-callbacks) refers to the documentation of the FSM package, but we've since implemented our own callback management. The order reported there is still correct, except the generic targets are not supported.

teo (changes requested):

docs/handbook/operation_order.md
knopers8: Thanks, I wasn't aware. I will update the doc.

teo (approved):

PR #575, opened 2024-05-31
[core] Convert ODC device and env states to ECS and publish

teo:
https://its.cern.ch/jira/browse/OCTRL-897

@knopers8 please review but we wait to merge until George is ready.

@graduta several changes to ODC-related payloads, possibly one of them breaking:

* The ODC-provided part of the response to `GetIntegratedServices` is now a map from env ID (partition ID) to a map including ODC and ECS state. Previously it used to be a map from env ID to ODC state string. The value for each env ID now looks like:
```
				"state":    partitionInfo.State,
				"ecsState": partitionInfo.EcsState,
```
* The ODC-provided part of the integrated services payload in the response to `GetEnvironments`: new field `ecsState` in each partition map (non-breaking)
* The ODC-provided part of the integrated services payload in the response to `GetEnvironment` and `NewEnvironment`: same as above, plus `devices` is now a uint64-keyed map (it used to be a list with no guaranteed order, so this is the possibly-breaking part). The `uint64` key corresponds to the `taskId` string within each `devices` entry. Furthermore, each `devices` entry now has an additional `ecsState` field.
* The `aliecs.inegrated_services.odc` topic now gets way more messages, in addition to the previous `odc.partitionStateChanged` message. The new kind of message is produced on every ODC-state change of an ODC device. Specifically, the new message has name `odc.deviceStateChanged` and its payload is the JSON representation of the following:
```
type deviceStateChangedEventPayload struct {
	PartitionId      uid.ID   `json:"partitionId"`
	DdsSessionId     string   `json:"ddsSessionId"`
	DdsSessionStatus string   `json:"ddsSessionStatus"`
	State            string   `json:"state"`
	EcsState         sm.State `json:"ecsState"`
	TaskId           string   `json:"taskId"`
	Path             string   `json:"path"`
	Ignored          bool     `json:"ignored"`
	Host             string   `json:"host"`
}
```


core/integration/odc/fairmq/fairmq.go
```diff
@@ -0,0 +1,80 @@
...
+ * immunities granted to it by virtue of its status as an
+ * Intergovernmental Organization or submit itself to any jurisdiction.
+ */
+
+package fairmq
+
+var (
+	fairMqStateMap map[string]string
+)
+
```
justonedev1, line 30: why this file does not have unit test? I think that it would be really helpful if there was exhaustive suite of tests, which describe all supported and expected mappings and don't rely on other people (probable George) to report that we made a stupid mistake in mapping.

core/integration/odc/plugin.go
```diff
@@ -248,44 +250,97 @@ func (p *Plugin) queryPartitionStatus() {
...
+						} else {
+							// if the state hasn't changed, we bail
+							continue
+						}
+					}
+
+					device.EcsState = fairmq.ToEcsState(device.State, oldState)
+
+					// since the odc-state of the task has changed, we must publish the event
+					payload := map[string]interface{}{
```
justonedev1, line 282: what is this payload? where is it defined? It seems to me that just in the middle of some file you are using "random" json payload. If there is a mistake it would be nightmare to figure out where is this coming from. Isn't it possible to create some payload struct with given members that will be translated correctly instead of using string map?

core/integration/odc/plugin.go
```diff
@@ -248,44 +250,97 @@ func (p *Plugin) queryPartitionStatus() {
...
 					log.WithField("level", infologger.IL_Support).
 						WithField("partition", id.String()).
 						WithField("oldState", existingPartition.State).
 						WithField("state", partitionInfo.State).
-						Warn("could not notify environment manager of ODC partition state change event")
+						Info("ODC Partition state changed")
+
+					partitionInfo.EcsState = fairmq.ToEcsState(partitionInfo.State, existingPartition.State)
+
+					payload := map[string]interface{}{
```
justonedev1, line 313: the same as before about payload. But here we are just sending less info. I really don't like using raw strings and different number of members for conveying similar message. I find it really error prone

justonedev1 (changes requested):
I don't know whether the fmq mapping is correct, that would be for @knopers8 ... but I found some things that I commented on

core/integration/odc/plugin.go
```diff
@@ -248,44 +250,97 @@ func (p *Plugin) queryPartitionStatus() {
...
+					the.EventWriterWithTopic(TOPIC).WriteEvent(&pb.Ev_IntegratedServiceEvent{
+						Name:          "odc.deviceStateChanged",
+						EnvironmentId: id.String(),
+						Payload:       string(payloadJson[:]),
 					})
-				} else {
+				}
+
+				// detection of env (ODC partition) state change + event publication
+				if existingPartition.State != partitionInfo.State {
```
knopers8, line 304:
```suggestion
				if oldPartitionInfo.State != partitionInfo.State {
```
could you use `old` for both device info and partition info? otherwise it's quite confusing to read if for some things it's `existing` and for others `old`.

It can be also `existing` for both.

knopers8 (changes requested):
Thanks! Please see my comments.

core/integration/odc/fairmq/fairmq.go
```diff
@@ -0,0 +1,80 @@
...
+		"INITIALIZED":              "STANDBY",
+		"BINDING":                  "STANDBY",
+		"BOUND":                    "STANDBY",
+		"CONNECTING":               "STANDBY",
+		"DEVICE READY CONFIGURING": "STANDBY", // needs special case
+		"INITIALIZING TASK":        "STANDBY",
+
+		"RESETTING TASK":         "CONFIGURED",
+		"DEVICE READY RESETTING": "CONFIGURED", // needs special case
+		"RESETTING DEVICE":       "CONFIGURED",
```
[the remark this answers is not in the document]
teo: Fair enough, and I can implement this, though I guess in that case instead of looking at the previous ODC state we should account for which ODC state changes are invariant with respect to the ECS state.

AliECS cannot affect ODC states with the granularity you describe, so for this to happen ODC would've had to reach `INITIALIZED` and then go back instead of forward on its own accord contrary to AliECS input. In any case we end up back in `IDLE` or possibly `ERROR` so not much changes for AliECS in between.

For this matter, then even the case of `IDLE`→`INITIALIZED`→`BOUND`→`DEVICE READY`→`RESETTING DEVICE` (similarly not requestable by AliECS) should also resolve to `STANDBY`, and in order to implement this we'd need to look back 2 states instead of 1 as we do now (or forget about ODC state lookback and take ECS state into account instead). When the interface was negotiated with ODC we explicitly demanded a single call to go from `IDLE` to `READY`, and for ODC to handle any mess that might happen in between.

core/integration/odc/plugin.go
teo: It's very much not random and its contents are purposefully chosen. It's defined here. Yes it's possible to turn it into a payload struct :+1:

knopers8 (approved):

justonedev1 (approved):

core/task/sm/state_test.go
```diff
@@ -0,0 +1,88 @@
+package sm
```
knopers8, line 1: in general it's better to do `git mv <oldfile> <newfile>` instead of plain `mv`, since this works better with `git blame` and alike

knopers8 (approved):

PR #581, opened 2024-06-26
[core] Limit ODC error string length

teo:
@graduta FYI


graduta:
Thank you for the quick action!

knopers8 (approved):

PR #582, opened 2024-06-27
[core] Improve handling of Mesos resource offers

teo:
[OCTRL-777](https://its.cern.ch/jira/browse/OCTRL-777)

This commit addresses an issue with timing between a Mesos REVIVE call and its corresponding OFFERS event.

Specifically:

* The channel that pipes incoming deployment requests into the OFFERS handler is now buffered.
* We retry an unsatisfiable task deployment 3 times before giving up.
* The deployment request is now passed as a pointer.
* The response from the OFFERS handler to task.Manager.acquireTasks now goes through its own channel, one per request.
* The deployment request is now enqueued immediately *before* a REVIVE, as opposed to after, in order to prevent a race with Mesos.

### DIFF
--- a/core/task/manager.go
+++ b/core/task/manager.go
@@ -75,6 +75,7 @@ type ResourceOffersOutcome struct {
 type ResourceOffersDeploymentRequest struct {
 	tasksToDeploy Descriptors
 	envId         uid.ID
+	outcomeCh     chan ResourceOffersOutcome
 }
 
 type Manager struct {
@@ -88,8 +89,7 @@ type Manager struct {
 	classes *taskclass.Classes
 	roster  *roster
 
-	resourceOffersDone <-chan ResourceOffersOutcome
-	tasksToDeploy      chan<- ResourceOffersDeploymentRequest
+	tasksToDeploy chan<- *ResourceOffersDeploymentRequest
 
 	reviveOffersTrg chan struct{}
 	cq              *controlcommands.CommandQueue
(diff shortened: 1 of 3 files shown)


core/task/manager.go
```diff
@@ -491,93 +490,107 @@ func (m *Manager) acquireTasks(envId uid.ID, taskDescriptors Descriptors) (err e
...
+			undeployableDescriptors = roOutcome.undeployable
+
+			log.WithField("tasks", deployedTasks).
+				WithField("partition", envId).
+				Debugf("resourceOffers is done, %d new tasks running", len(deployedTasks))
+
+			if len(deployedTasks) != len(tasksToRun) {
+				// ↑ Not all roles could be deployed. If some were critical,
+				//   we cannot proceed with running this environment. Either way,
+				//   we keep the roles running since they might be useful in the future.
```
knopers8, line 540: I don't understand. Where is the part of code which tries to re-deploy only the undeployed tasks?

knopers8 (approved):
I have to admit I don't fully understand the code here, but I would not block the PR to have it in the next release and rather discuss it during our ECS weekly.

core/task/manager.go
teo: I guess the comment is misleading. They would be useful in the future if task reuse was enabled, but it was disabled some years ago since start-stop-start wasn't there yet. In case of partial deployment we'd actually fail, but to my knowledge the current undeployable cases we have are actually early failures where no tasks are deployed.

PR #583, opened 2024-06-27
[core] test environment's FSM, handling hooks and fix discovered issues

knopers8:


core/environment/fsm_test.go
```diff
@@ -0,0 +1,288 @@
...
+
+			err := env.Sm.Event(context.Background(), "CONFIGURE", NewDummyTransition("CONFIGURE"))
+
+			Expect(err).NotTo(HaveOccurred())
+			v, ok := env.workflow.GetUserVars().Get("root.call_called")
+			Expect(ok).To(BeTrue())
+			Expect(v).To(Equal("true"))
+		})
+
+		It("should return an error if a critical hook times out", func(ctx SpecContext) {
```
knopers8, line 261: @teo could please cross-check this case and tell me if I am missing something here? It fails for some reason - the call hangs infinitely and the timeout mechanism does not kick in (I added a `SpecTimeout` to avoid having CI hang).

core/environment/fsm_test.go
```diff
@@ -0,0 +1,288 @@
...
+				Expect(env.Sm.Cannot("EXIT")).To(BeTrue())
+			})
+		})
+		When("FSM is in ERROR", func() {
+			It("should allow for RECOVER transition", func() {
+				env.Sm.SetState("ERROR")
+				Expect(env.Sm.Can("RECOVER")).To(BeTrue())
+				// fixme: is this correct that we do not allow for EXIT?
+				//  in principle, in TeardownEnvironment we ignore the FSM and do it ourselves,
+				//  but I don't see a reason not to use allow for it
```
teo, line 161: An arbitrary choice I think. Currently we have no ERROR recovery, and this FSM maps to task states. But killing the task is possible even when that task's FSM disobeys or doesn't work.

core/environment/fsm_test.go
```diff
@@ -0,0 +1,288 @@
...
+				// fixme: shouldn't we add also DESTROY and TEARDOWN in FSM?
+				//  is even DESTROY different from TEARDOWN?
```
teo, line 163: They seem to be used as TEARDOWN→DESTROY but it's all fake anyway, they represent transitioning to the exit state if possible, and then even if not, killing all tasks.

teo (approved):

PR #588, opened 2024-07-09
[core] Set run timestamps before executing triggers with weight 0

knopers8:
I still want to test it with TRG plugin enabled, but otherwise, it seems already OK.

### DIFF
--- a/core/environment/environment.go
+++ b/core/environment/environment.go
@@ -172,6 +172,22 @@ func newEnvironment(userVars map[string]string, newId uid.ID) (env *Environment,
 					Message:        "transition step starting",
 				})
 
+				// first, we execute hooks which should be executed before an event officially starts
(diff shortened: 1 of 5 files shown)


knopers8:
It has to come with https://github.com/AliceO2Group/ControlWorkflows/pull/655

teo (dismissed):

knopers8:
I tested it on staging with DCS, TRG, ODC, CCDB, BKP, DD, Kafka plugins enabled with a SYNTH 650kHz pp configuration. EMC, CPV, PHS, TRD detectors had to be excluded to allow for trigger to start.

I confirmed that the order of plugin execution remained the same.

I had to add one fix (delete run number after positive STOP_ACTIVITY hooks are called), which allowed CCDB RunStop to still work correctly. Otherwise, all the plugin logs remained the same as before the changes.

It's good to go as far as I am concerned.

teo (approved):

PR #590, opened 2024-07-10
OCTRL-899 [core] use SOSOR and EOEOR as run duration for GRP ECS object

knopers8:
Ruben and Chiara need the O2 run duration in GRP ECS objects to be the narrowest timespan which includes CTP start and stop. Also start of run time should be ready available when the tasks are started. Because of these two reasons, we should use SOSOR and EOEOR as the timestamps in GRP ECS object.


knopers8:
Keeping as WIP until https://github.com/AliceO2Group/Control/pull/588 is merged.

teo (approved):

knopers8:
I tested it on staging, it seems to be working as expected as long as https://github.com/AliceO2Group/ControlWorkflows/pull/656 is also included.

PR #595, opened 2024-07-17
added logging filters for IL

justonedev1:
I had to change how directhook is created, because when it is triggered from `init()` it just uses default values all the time and it isn't settable.  So I moved it from `init()` to `main()` after config is being loaded and parsed.

### DIFF
--- a/common/logger/infologger/directhook.go
+++ b/common/logger/infologger/directhook.go
@@ -38,6 +38,7 @@ import (
 
 	"github.com/AliceO2Group/Control/common/utils"
 	"github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
 )
 
 const INFOLOGGER_MAX_MESSAGE_SIZE = 1024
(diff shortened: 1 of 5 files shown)


common/logger/infologger/directhook.go
```diff
@@ -38,6 +38,7 @@ import (
 
 	"github.com/AliceO2Group/Control/common/utils"
 	"github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
```
teo, line 41: The inconvenience here is that you've just pulled viper as a dependency of the executor by way of `DirectHook`, and the executor doesn't even use Viper. I don't know how fat Viper is in absolute terms, but it's a fairly comprehensive library with over a dozen of its own dependencies. So far I've mostly tried to keep the executor small and light as much as possible to ensure fast startup (and even with that, it's up to 28MB now). Most likely it's not a big deal, but it would be good if you could check the effect of this change on binary size, and if significant, try to find a way to avoid the added dependency.

teo (changes requested):

common/logger/infologger/directhook.go
teo: Actually, looks like I did that recently by using `utils.TimeTrack` in the executor (and only this so far, AFAICT), so a Viper dependency is already there, unfortunately. I think it would still be good to keep Viper out of the executor, but it's probably out of scope of this logging effort.

teo (approved):

PR #600, opened 2024-07-30
[core] OCTRL-911 Transitions should not be performed concurrently

knopers8:
Tested on staging with a few different cases when tasks go to ERROR, the protection works and there are no deadlocks.


teo (dismissed):

knopers8:
Do not merge yet, I managed to cause a segfault with two concurrent DESTROY attempts. While the 2nd one waits until the 1st one succeeds, it sees a `nil` environment and it crashes. I have to fix that first.

knopers8:
OK, dumb me, it's good now.

PR #602, opened 2024-08-02
[core] OCTRL-912 Cancel calls pending await upon environment destruction

knopers8:
This addressed the problem of having stuck calls waiting for something to receive from c.await which outlive the parent environments. This may happen if awaited trigger never occurs after it has been started, e.g. if trigger is before_START_ACTIVITY and await is after_START_ACTIVITY, but the environments fails to transition to RUNNING.


knopers8:
@teo @justonedev1 could you review please?

teo (approved):

PR #603, opened 2024-08-22
Refactor gera.Map + fix KV override with empty value issue + unit tests

teo:
https://its.cern.ch/jira/browse/OCTRL-916


common/gera/map_test.go
```diff
@@ -0,0 +1,626 @@
...
+		It("should wrap correctly", func() {
+			wrappedDefaults = defaults2.Wrap(defaults1)
+			Expect(wrappedDefaults).NotTo(BeNil())
+			Expect(wrappedDefaults).To(Equal(defaults2))
+
+			wrappedVars = vars2.Wrap(vars1)
+			Expect(wrappedVars).NotTo(BeNil())
+			Expect(wrappedVars).To(Equal(vars2))
+
+			wrappedUserVars = userVars1
```
knopers8, line 325: A leftover?

common/gera/map_test.go
```diff
@@ -0,0 +1,626 @@
...
+			Expect(unwrapped).NotTo(BeNil())
+			Expect(unwrapped).To(Equal(vars1))
+
+			wrappedVars = vars2.Wrap(vars1)
+			Expect(wrappedVars).NotTo(BeNil())
+		})
+
+		var flattenedDefaults, flattenedVars, flattenedUserVars, flattenedAll map[string]string
+
+		It("should flatten correctly", func() {
```
knopers8, line 339: By `Flatten()`, do you mean to remove the nesting by representing the whole path to a value in the key? If so, could we perhaps test this? I don't quite understand what do we test now.

knopers8 (commented):
Stupid question: what does "gera" stand for?

core/workflow/rolebase.go
```diff
@@ -56,12 +57,42 @@ type roleBase struct {
...
-	Enabled  string              `yaml:"enabled,omitempty"`
+	Defaults *gera.WrapMap[string, string] `yaml:"defaults,omitempty"`
+	Vars     *gera.WrapMap[string, string] `yaml:"vars,omitempty"`
+	UserVars *gera.WrapMap[string, string] `yaml:"-"`
+	Locals   map[string]string             `yaml:"-"` // only used for passing iterator from template to new role
+	Bind     []channel.Inbound             `yaml:"bind,omitempty"`
+	Enabled  string                        `yaml:"enabled,omitempty"`
+}
+
+func kvStoreUnmarshalYAMLWithTags(w gera.Map[string, string], unmarshal func(interface{}) error) error {
```
justonedev1, line 68: shouldn't this function be also tested?

common/gera/map.go
```diff
@@ -29,79 +29,130 @@
...
 
 import (
+	"sync"
+
 	"dario.cat/mergo"
 )
 
-type Map interface {
-	Wrap(m Map) Map
+type Map[K comparable, V any] interface {
```
justonedev1, line 37: is there a reason why use type arguments? I see only Map[string,string] used everywhere.

common/gera/map_test.go
```diff
@@ -0,0 +1,626 @@
...
+import (
+	"maps"
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"gopkg.in/yaml.v3"
+)
+
+var _ = Describe("hierarchical key-value store", func() {
```
justonedev1, line 36: Is there some code coverage report generated? It would be nice to see whether you really test everything, because I cannot tell if everything is tested, from simple read through

justonedev1 (changes requested):

teo:
> Stupid question: what does "gera" stand for?

Not a stupid question at all. Back in the day I wanted a short package name (still do), which would have been `hiera`, but I didn't want to cause future confusion with the Hiera KV store used by Puppet, so instead of the Latin root for "hierarchy" I went with the Italian one where "hi" becomes a soft "g".

common/gera/map_test.go
teo: More like consistency. It's used further down.

common/gera/map_test.go
teo: No, flattening doesn't modify keys. It just merges a hierarchy of wrapped maps into a single `map[K]V` while respecting the override order.

core/workflow/rolebase.go
teo: We can indeed test it, but a test probably belongs in the `workflow` package test and not in the one for the `gera` package.

common/gera/map.go
teo: It would probably be safe to make the key a string forever, not so sure about the value. Basically `Map` already existed and I didn't want to preclude using non-string values in the future, if we so wish, mostly for the purposes of the template system, as well as using `gera` in contexts other than the template system. Since Go got generics a few years ago, the untemplated `gera.Map` became old style go, relying on `interface{}`, so the paths forward were either (1) to remove it and only use `gera.StringMap`, or (2) to update `gera.Map` for modern Go. I choose 2.

common/gera/map_test.go
teo: `ginkgo -r -v -race --trace --coverpkg=github.com/AliceO2Group/Control/common/gera --coverprofile=.coverage-report.out`
...
```
Ran 21 of 21 Specs in 0.023 seconds
SUCCESS! -- 21 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
coverage: 80.2% of statements in github.com/AliceO2Group/Control/common/gera
composite coverage: 80.2% of statements
```

teo: Or actually, just `make coverage`. AFAICT you added the command this April, and thanks for that!

common/gera/map.go
justonedev1: okay, thank you

common/gera/map_test.go
justonedev1: oh heck... you are right, I already forgot about it :laughing:  I took a look at the coverage and it looks fine. However, I did not see any error path tests. But I trust you, that it is not necessary.

common/gera/map_test.go
knopers8: thanks, clear

knopers8:
> Not a stupid question at all. Back in the day I wanted a short package name (still do), which would have been `hiera`, but I didn't want to cause future confusion with the Hiera KV store used by Puppet, so instead of the ~Latin~ ancient Greek root for "hierarchy" I went with the Italian one where "hi" becomes a soft "g".

Thanks, that explains it. Please add a word about it somewhere, e.g. at the top of `common/gera/map.go`, then it's good with me.

knopers8 (dismissed):

justonedev1 (dismissed):

knopers8 (approved):

PR #605, opened 2024-09-02
[core] Emit environment events from CreateAutoEnvironment

teo:


knopers8 (approved):
Huh, I thought BKP already had moved the de-duplicated call.

teo:
> Huh, I thought BKP already had moved the de-duplicated call.

It's not on BK to do so, they can only receive events through Kafka. It's on the GUI to stop calling `NewAutoEnvironment` and consuming the gRPC response stream in favour of calling `NewEnvironmentAsync` and consuming from Kafka.

PR #606, opened 2024-09-02
Add support for IsTruthy/IsFalsy

teo:


docs/handbook/configuration.md
```diff
@@ -18,7 +167,7 @@ roles:
 
 In the absence of an explicit `critical` trait for a given task role, the assumed default value is `critical: true`.
 
-## State machine callbacks
+#### State machine callbacks moments
```
knopers8, line 170: I'm not sure what to propose, but perhaps there is a better place for this subsection, since it relates both to hook tasks and calls.

knopers8 (approved):
Many thanks for the extended documentation!

PR #607, opened 2024-09-04
Add nanosecond timestamp + workflow info to environment events

teo:
This PR does 2 things:

* [OCTRL-917](https://its.cern.ch/jira/browse/OCTRL-917) - adds a new `WorkflowTemplateInfo` field to all `EnvironmentEvent`s pushed to Kafka, that includes the public trait of the workflow template as requested, plus the name, description and workflow path.
* [OCTRL-918](https://its.cern.ch/jira/browse/OCTRL-918) - in addition to the millisecond timestamp, adds a nanosecond timestamp field `timestampNano` to all events pushed to Kafka. The timestamp itself is identical to the existing ms timestamp, it is just outputted with higher precision.

@martinboulais @graduta


core/environment/utils.go
```diff
@@ -28,38 +28,47 @@ import (
...
-	"gopkg.in/yaml.v3"
 	"os"
 	"sort"
+
+	"github.com/AliceO2Group/Control/core/the"
+	"gopkg.in/yaml.v3"
 )
 
-func parseWorkflowPublicInfo(workflowExpr string) (bool, string, error) {
+type WorkflowPublicInfo struct {
```
knopers8, line 38:
```suggestion
type WorkflowInfo struct {
```
the name could be updated.

knopers8 (approved):
good for me, but I will wait for Martin or George's confirmation before merging.

martinboulais:
@knopers8 So if I understood correctly, if `event.environmentEvent.workflowTemplateInfo.public` is false, it should not be stored in Bookeeping?

knopers8:
@martinboulais I would even go further, I would require `event.environmentEvent.workflowTemplateInfo.name == "readout-dataflow"`. In the ECS GUI you can notice also the template `qc-server-workflow` and as far as I know, it also should not be recognized by BKP (and historically was not).

martinboulais:
This seems a bit too hacky to me, no? I don't really like the hardcoded values like this, especially when it's from another service.
Why not simply marking the `public` to false in the workflow template?

knopers8:
What I think would be the best is to add another field/variable in the templates such as `bookkeeped` and  `WorkflowInfo` would include it.

vascobarroso:
@knopers8 @martinboulais I think the suggestion to add a new field makes a lot of sense. The current `WorkflowPublicInfo` value is to let the AliECS GUI know if it should be shown or not in the env list. Storing in bookkeeping is similar but not fully equivalent, so a new `WorkflowStoreInBookkeeping` or something like that makes sense.

martinboulais:
@vascobarroso @knopers8 I agree that adding a new field makes quite some sense to me.

core/environment/utils.go
knopers8: ah ok, i think i understand what "public" refers to now.

teo:
@knopers8 `bookkept` is now a variable as discussed with @martinboulais, but the rest of this PR still applies.

knopers8 (approved):
thank you

PR #608, opened 2024-09-06
AgentCache thread safety

justonedev1:


knopers8 (approved):

knopers8:
How did you notice it? Did you see a crash associated with this function?

justonedev1:
I was once again going through the code and looking for something that might explain double executor problem... this is not it, but it is a bug, which probably never manifested

PR #609, opened 2024-09-13
OCTRL-920 Fixes for stuck calibration runs 

knopers8:
The three commits fix what was observed as wrong while investigating the case with env `2pedNJwSjL8`. The updated core was confirmed to work correctly in this test scenario.

### DIFF
--- a/core/task/manager.go
+++ b/core/task/manager.go
@@ -938,6 +938,7 @@ func (m *Manager) updateTaskState(taskId string, state string) {
 	taskPtr := m.roster.getByTaskId(taskId)
 	if taskPtr == nil {
 		log.WithField("taskId", taskId).
+			WithField("state", state).
 			Warn("attempted state update of task not in roster")
 		return
 	}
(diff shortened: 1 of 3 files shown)


core/task/manager.go
```diff
@@ -1052,20 +1054,27 @@ func (m *Manager) KillTasks(taskIds []string) (killed Tasks, running Tasks, err
...
+	}
 	// TODO: use grouping instead of 2 passes of filtering for performance
 	toKill := m.roster.filtered(taskCanBeKilledFilter)
-	unkillable := m.roster.filtered(func(t *Task) bool { return !taskCanBeKilledFilter(t) })
 
 	if len(toKill) < len(taskIds) {
+		unkillable := m.roster.filtered(func(t *Task) bool { return !taskCanBeKilledFilter(t) })
 		log.WithField("taskIds", strings.Join(unkillable.GetTaskIds(), ", ")).
-			Debugf("some tasks cannot be physically killed (already dead?), will instead only be removed from roster")
+			Debugf("some tasks cannot be physically killed (already dead or being killed in another goroutine?), will instead only be removed from roster")
```
justonedev1, line 1068: does this happen a lot? Shouldn't this be a Warning?

knopers8: i think this can be a somewhat valid situation if two kill requests happen subsequently, so I'm not sure if a warning is needed. But I don't know how often this may happen.

justonedev1 (approved):

PR #613, opened 2024-09-18
QC-779  Multiple mesos executor fix

justonedev1:
sorry for some of the formatting fixes.
There was a race condition which was in method `makeTaskForMesosResources` where we sharing `global` object to write. This should create a deep copy of given object and it seems to fix the problem on staging.

### DIFF
--- a/core/task/schedulerstate.go
+++ b/core/task/schedulerstate.go
@@ -36,6 +36,7 @@ import (
 	"github.com/AliceO2Group/Control/common/logger/infologger"
 	"github.com/AliceO2Group/Control/core/controlcommands"
 	"github.com/AliceO2Group/Control/core/task/schedutil"
+	"github.com/gogo/protobuf/proto"
 	"github.com/looplab/fsm"
 	mesos "github.com/mesos/mesos-go/api/v1/lib"
 	"github.com/mesos/mesos-go/api/v1/lib/backoff"
@@ -214,17 +215,6 @@ func (state *schedulerState) Start(ctx context.Context) {
 	}()
 }
 
-func (state *schedulerState) CopyExecutor() *mesos.ExecutorInfo {
-	executorInfoCopy := &mesos.ExecutorInfo{}
-
-	marshaled, err := state.executor.Marshal()
-	if err != nil {
-		return nil
-	}
-
-	err = executorInfoCopy.Unmarshal(marshaled)
-	if err != nil {
-		return nil
-	}
-	return executorInfoCopy
+func (state *schedulerState) CopyExecutorInfo() *mesos.ExecutorInfo {
+	return proto.Clone(state.executor).(*mesos.ExecutorInfo)
 }
--- a/core/task/schedulerstate.go
+++ b/core/task/schedulerstate.go
@@ -94,7 +94,6 @@ func NewScheduler(taskman *Manager, fidStore store.Singleton, shutdown func()) (
 			viper.GetFloat64("executorMemory")),
 		viper.GetDuration("mesosJobRestartDelay"),
 	)
-
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


core/task/schedulerstate.go
```diff
@@ -208,10 +206,25 @@ func (state *schedulerState) Start(ctx context.Context) {
...
-			state.sm.Event(context.Background(), "GO_ERROR", err) //TODO: use error information in GO_ERROR
+			state.sm.Event(context.Background(), "GO_ERROR", err) // TODO: use error information in GO_ERROR
 		} else {
 			log.Debug("scheduler quit, no errors")
 			state.sm.Event(context.Background(), "EXIT")
 		}
 	}()
 }
+
+func (state *schedulerState) CopyExecutor() *mesos.ExecutorInfo {
```
knopers8, line 217: can you rename to `CopyExecutorInfo` for clarity?

also, perhaps [`proto.Clone()`](https://pkg.go.dev/google.golang.org/protobuf/proto#Clone) might be faster than marshalling and unmarshalling?

justonedev1: The reason why I did not use `Clone` originally was, that it did not build, hence the stupid marshalling/unmarshalling. I got error 
```
*mesos.ExecutorInfo does not implement protoreflect.ProtoMessage (missing method ProtoReflect)
but the object in defined in filename.pb.go
```
but now it works... ffs literally the same code I wonder what I did differently before (I even copied the same code as before :D)

justonedev1:
Nope, this was the only place where the executor from state is written to. It is supposed to be this immutable object that is safe to read from, so it is just created as a part of `schedulerState` [here](https://github.com/AliceO2Group/Control/blob/master/core/task/schedulerstate.go#L122) and the `executorInfo` is created few lines above [here](https://github.com/AliceO2Group/Control/blob/master/core/task/schedulerstate.go#L90) ... So nothing now modifies original `executorInfo`, just the copies I create.

justonedev1:
And one more thing... this is a problem of golang, as there is no const keyword like in eg. c++, rust, ... If this was correctly written c++ code/ rust code this would not be able to happen.

knopers8:
> Nope, this was the only place where the executor from state is written to. It is supposed to be this immutable object that is safe to read from, so it is just created as a part of `schedulerState` [here](https://github.com/AliceO2Group/Control/blob/master/core/task/schedulerstate.go#L122) and the `executorInfo` is created few lines above [here](https://github.com/AliceO2Group/Control/blob/master/core/task/schedulerstate.go#L90) ... So nothing now modifies original `executorInfo`, just the copies I create.

Indeed, I missed the fact that executor ID is set just a line after the (now) deep-copy:
```
	executor := state.CopyExecutor()
	executor.ExecutorID.Value = taskPtr.GetExecutorId()
```
before I thought that the right executor ID is already in `state.executor` when we receive it.

all clear, thanks!

core/task/schedulerstate.go
knopers8: it could have been that you imported a wrong lib, I saw in GoLand that you could pick three different libraries with `proto.Clone`

core/task/schedulerstate.go
```diff
@@ -36,6 +36,7 @@ import (
 	"github.com/AliceO2Group/Control/common/logger/infologger"
 	"github.com/AliceO2Group/Control/core/controlcommands"
 	"github.com/AliceO2Group/Control/core/task/schedutil"
+	"github.com/gogo/protobuf/proto"
```
knopers8, line 39: [`gogo/protobuf`](https://github.com/gogo/protobuf) is deprecated though, can you try using a different one?

core/task/schedulerstate.go
justonedev1: maybe... truth is, that I let lsp handle adding dependencies...

core/task/schedulerstate.go
knopers8: i think this is the official one now: https://github.com/golang/protobuf

justonedev1: no, it will not work, see previous build error.. but this package is used in mesos, so we are using it either way inherently, so I don't mind using it for cloning an object, that is already created by this package from mesos

justonedev1: but I personally also don't mind using marshall and unmarshall because we are using it once per run and per offer

knopers8: ok, since mesos brings it anyway, let's keep gogo/protobuf, thanks for checking.

knopers8 (approved):

PR #615, opened 2024-09-20
removing kafka event in DCS plugin for EOF received from DCS

justonedev1:
The reason for this is described in ticket [OCTRL-924](https://its.cern.ch/jira/browse/OCTRL-924)

### DIFF
--- a/core/integration/dcs/plugin.go
+++ b/core/integration/dcs/plugin.go
@@ -1041,7 +1041,6 @@ func (p *Plugin) CallStack(data interface{}) (stack map[string]interface{}) {
 			})
 		}
 		return
-
 	}
 	stack["StartOfRun"] = func() (out string) { // must formally return string even when we return nothing
 		var err error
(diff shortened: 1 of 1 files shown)


knopers8 (approved):

core/integration/dcs/plugin.go
```diff
@@ -1416,17 +1415,6 @@ func (p *Plugin) CallStack(data interface{}) (stack map[string]interface{}) {
...
-					Name:                call.GetName(),
-					OperationName:       call.Func,
-					OperationStatus:     pb.OpStatus_ONGOING,
-					OperationStep:       "perform DCS call: StartOfRun",
-					OperationStepStatus: pb.OpStatus_DONE_ERROR,
-					EnvironmentId:       envId,
-					Payload:             string(payloadJson[:]),
-					Error:               logMsg,
-				})
-
```
teo, line 1429: I agree with this change. Beyond the question of whether the stream `io.EOF` was timely or not from the DCS side, it could be argued that a stream being closed either way should not generate a Kafka event, because it concerns a technicality of DCS-ECS communication and says nothing about the actual state of the DCS.

However, the behaviour should be consistent across all DCS operations. @knopers8 could you please make sure this change also applies to PFR and EOR?

justonedev1: sure, I will do so.

PR #616, opened 2024-09-25
[OCTRL-928] Add kafka event for DCS SOR state SOR_PROGRESSING

justonedev1:
Added kafka message to DCS for any detector change for any detector


knopers8 (approved):
I agree about the boilerplate... I'm not sure what to do about it exactly, but let's think about it.

PR #622, opened 2024-10-09
[core] Include state in DCS op failure events

teo:


teo:
@vascobarroso @graduta FYI

knopers8 (dismissed):

PR #624, opened 2024-10-18
ddsched plugin gRPC calls have timeout by default

justonedev1:
I did this as a proposal, how to add gRPC timeout to all calls in one plugin by default. However I am not sure whether this is desirable, because I found multiple TIMEOUTS in ddsched plugin, some of them are not even used. So I would like to ask @teo what do you think about this.


teo (changes requested):

teo:
> I did this as a proposal, how to add gRPC timeout to all calls in one plugin by default. However I am not sure whether this is desirable, because I found multiple TIMEOUTS in ddsched plugin, some of them are not even used. So I would like to ask @teo what do you think about this.

I agree with this in principle. `ddsched` is specific because we have to poll for op completion, but the client-side timeout is still missing. Concerning the unused timeout constants, it could be they were planned but never materialized, or were used in a previous version (likely the former).

teo (approved):

PR #625, opened 2024-10-19
reset deploymentSuccess variable in between retries

justonedev1:
@teo please can you confirm, that cause of OCTRL-777 could be this stupid?


justonedev1:
Please don't merge it immediately

justonedev1:
This was replaced by OCTRL-777. Closing

PR #627, opened 2024-10-21
[OCTRL-777] Environment declared as failed during the deployment

justonedev1:
I was able to trigger "no resources" error on my openstack setup. The problem was that variables were not reset correctly when we were deploying and failing in retry loop. Moreover I moved reporting of UNDEPLOYABLE status after whole retry loop is done. With these changes we are able to successfully retry deployment, if we fail for the first time without involvement of shifter.

Moreover I improved logging a bit, I sent also the error reports to Roberto whether he is fine with the amount of errors reported in this case + the messaging. So please don't merge this yet.


core/task/manager.go
```diff
@@ -326,6 +326,14 @@ func (m *Manager) RefreshClasses(taskClassesRequired []string) (err error) {
 	return
 }
 
+// prefix will be prepended before name of descriptor
+func logDescriptors(prefix string, logFunc func(format string, args ...interface{}), descriptos Descriptors) {
+	for _, desc := range descriptos {
```
teo, line 331: `descriptors`

teo (dismissed):

teo (approved):

PR #630, opened 2024-10-23
[core] Ensure TRG timeouts are obeyed + add polling timeout

teo:


core/integration/trg/plugin.go
```diff
@@ -853,6 +866,10 @@ func (p *Plugin) CallStack(data interface{}) (stack map[string]interface{}) {
 			return
 		}
 
+		timeout := callable.AcquireTimeout(TRG_START_TIMEOUT, varStack, "RunStart", envId)
+		ctx, cancel := context.WithTimeout(context.Background(), timeout)
+		defer cancel()
+
```
justonedev1, line 872: why are you creating the context with timeout here and not next to the place where it is needed? (line 889)

justonedev1 (commented):
There are a lot of new timeouts, so I hope that you have verified that adding them will not break anything by timing out sooner than expected.

teo:
> There are a lot of new timeouts, so I hope that you have verified that adding them will not break anything by timing out sooner than expected.

This is my concern too, breaking existing behaviour, but I've been told the timeouts already set in readout-dataflow (which I also used as defaults, so `5s`) are realistic.

core/integration/trg/plugin.go
teo: I just did it in the same spot as in `ddsched`. Shouldn't change much either way, as between here and the call we only have declarations and a Kafka event.

PR #631, opened 2024-10-29
moved logs connected to unknown undeployable to Devel

justonedev1:
I talked with Vasco and it is possible to release and deploy before HI if we do just log changes. So I went over logs from OCTRL-777 logs and moved these to Devel level so crew will not get overwhelmed when this error (which is covered by retry) is triggered.

For the record: this PR is not meant to be as a whole solution for OCTRL-945, I am still working on more concise logging.


knopers8:
If we move those to DEVEL, are we left with some high-level error for an operator?

teo (dismissed):

justonedev1:
we I added more logs to Devel, and one general error that shift crew is going to see only when whole retry mechanism fails.

teo (approved):

PR #632, opened 2024-10-30
[core] Pass to FairMQ tasks abstract namespace endpoint paths

teo:


knopers8:
I remember that during the last weekly meeting you reported that this approach was not working. Could you explain what was the issue?

Also, could you please document in the commit description what is the motivation here or include the Jira ticket ID?

teo:
> I remember that during the last weekly meeting you reported that this approach was not working. Could you explain what was the issue?
> 
> Also, could you please document in the commit description what is the motivation here or include the Jira ticket ID?

Yes, sure. The issue was that FairMQ requires a specific syntax to express an abstract namespace socket name. Giulio advised me on this syntax. I've tested it on staging and it works.

The motivation is https://its.cern.ch/jira/browse/O2-5457

knopers8:
> Yes, sure. The issue was that FairMQ requires a specific syntax to express an abstract namespace socket name. Giulio advised me on this syntax. I've tested it on staging and it works.

Aha. In the meeting I understood it did not work.

> The motivation is https://its.cern.ch/jira/browse/O2-5457

Thank you. Can you please put the "O2-5457" in the commit description? It doesn't have to be the title, but just anywhere so it is trackable via `git blame`?

knopers8 (approved):

knopers8 (approved):

PR #633, opened 2024-10-30
Format whole codebase

justonedev1:
I formated everything with `gofmt -s -w .` from alidist environment golang, so other PR's not polluted by randomly reformatting whole gofile.


teo:
I would prefer to let our IDEs format as we go instead of this `gofmt` pass on everything, but I let @knopers8 decide.

justonedev1:
sorry, but no. You prefer that there are format changes in every PR? Look here https://github.com/AliceO2Group/Control/pull/632/files ? One line change causes whole file to be reformated, so I need to concentrate a bit more to figure out which changes are really relevant and which are just formatting  It might be possible to create github hook that reformats the PR automatically, but having format changes on almost every PR is absolutely no go. Plus `go fmt` or `gofmt` are meant to be used by golang developers.

knopers8 (approved):
Personally, I never proposed myself to do it to avoid obfuscating `git blame`, but I am not fundamentally against. Let's do it.

teo:
> sorry, but no. You prefer that there are format changes in every PR? Look here https://github.com/AliceO2Group/Control/pull/632/files ? One line change causes whole file to be reformated, so I need to concentrate a bit more to figure out which changes are really relevant and which are just formatting It might be possible to create github hook that reformats the PR automatically, but having format changes on almost every PR is absolutely no go. Plus `go fmt` or `gofmt` are meant to be used by golang developers.

To be clear I don't object to `go fmt`, it's what the IDE uses as well and it's the only reasonable way to ever format Go code. I simply don't mind format changes in every PR, I am totally ok with them and I think they are better than obfuscating `git blame`.

knopers8:
Merging then.

PR #637, opened 2024-11-27
Improve role query command documentation and fix proto docs build

teo:


knopers8 (approved):
thank you!

coconut/doc/coconut_configuration_show.md
```diff
@@ -29,6 +32,7 @@ coconut conf show <component> <entry> -s -e '{"key1": "value1", "key2": "value2"
   -l, --role string                      request configuration for this O² machine role
   -r, --runtype string                   request configuration for this run type (e.g. PHYSICS, TECHNICAL, etc.)
   -s, --simulate                         simulate runtime template processing on the configuration payload
+  -t, --timestamp string                 request configuration for this timestamp
```
knopers8, line 35: well, actually... how come this reappeared? I though i removed timestamp support for good...

teo: I think so too, I have no idea. Maybe we missed it in the show command?

PR #641, opened 2024-12-09
[core] Adding http metrics endpoint

justonedev1:
This PR is adding http endpoint to be scraped by other tools (eg. Telegraf). Apart from mechanism to create metrics endpoint I also added mechanism to publish [golang internal metrics](https://pkg.go.dev/runtime/metrics).

P.S.: I changed this to a draft, so we can discuss it before merging.

### DIFF
--- a/common/ecsmetrics/metrics.go
+++ b/common/ecsmetrics/metrics.go
@@ -1,0 +1,73 @@
+package ecsmetrics
+
+import (
+	internalmetrics "runtime/metrics"
+	"time"
+
+	"github.com/AliceO2Group/Control/common/logger"
+	"github.com/AliceO2Group/Control/common/monitoring"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	endRequestChannel chan struct{}
+	log               = logger.New(logrus.StandardLogger(), "ecsmetrics")
+)
+
+func gather() monitoring.Metric {
+	samples := []internalmetrics.Sample{
+		{Name: "/gc/cycles/total:gc-cycles"},
+		{Name: "/memory/classes/other:bytes"},
+		{Name: "/memory/classes/total:bytes"},
+		{Name: "/sched/goroutines:goroutines"},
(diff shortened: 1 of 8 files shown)


common/ecsmetrics/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+		switch sample.Value.Kind() {
+		case internalmetrics.KindUint64:
+			metric.AddValue(sample.Name, sample.Value.Uint64())
+		case internalmetrics.KindFloat64:
+			metric.AddValue(sample.Name, sample.Value.Float64())
+		case internalmetrics.KindFloat64Histogram:
+			fmt.Printf("Error: Histogram is not supported yet for metric [%s]", sample.Name)
+			continue
+		default:
+			fmt.Printf("Unsupported kind %v for metric %s\n", sample.Value.Kind(), sample.Name)
```
teo, line 44: Why `fmt` instead of our logger?

teo (dismissed):

common/ecsmetrics/metrics.go
justonedev1: because at first, I thought that I'd put this into metrics repo, I will change it to our logger

core/core.go
```diff
@@ -99,6 +143,11 @@ func Run() error {
 
 	// Plugins need to start after taskman is running, because taskman provides the FID
 	integration.PluginsInstance().InitAll(state.taskman.GetFrameworkID())
+	runMetrics()
+	defer ecsmetrics.StopGolangMetrics()
+	defer monitoring.Stop()
+
+	log.Infof("Everything initiated Listening on control port: %d", viper.GetInt("controlPort"))
```
knopers8, line 150: i don't think operators care about which ports are being listened on. imho, you could consider setting it to info/support.

knopers8 (approved):
thanks!

knopers8 (approved):

PR #650, opened 2025-01-15
[core] improved error reporting during env deployment/configuration

knopers8:
- "task '/opt/o2/bin/o2-readout-exe' on alio2-cr1-mvs11 (id 2sBwA3Z8yWU, name alio2-cr1-hv-gw01.cern.ch:/opt/git/ControlWorkflows/tasks/readout@12b11ac4bb652e1835e3e94806a688c951691d5f#2sBwA3Z8yWU) failed with error..." displayed by COG becomes "task 'readout' on alio2-cr1-mvs11 (id 2sBwA3Z8yWU) failed with error...". I believe the removed information is not useful for typical operators. Details can be always found in logs.
- "MesosCommand MesosCommand_Transition timed out for task 2sBwYFZ82wn" becomes "MesosCommand_Transition timed out for task 2sBwYFZ82wn" because by convention, all MesosCommands have "MesosCommand" in their names.
- If a transition request times out, more accurate error is reported in COG. "nil response" becomes "CONFIGURE could not complete for critical tasks, errors: task 'readout' on alio2-cr1-mvs11 (id 2sBwA3Z8yWU) failed with error: MesosCommand_Transition timed out for task 2sBwA3Z8yWU".
- In case a task transition times out, any other errors which happened at the same time (e.g. a task crashed during transition) are not omitted anymore in COG.
- "nil response" in task/manager.go becomes "no response from Mesos to CONFIGURE transition request within 120s timeout", but it will not be typically printed for tasks timing out during a transition.
- "nil response" in commandqueue.go becomes "did not receive neither response nor error for MesosCommand_Transition"

One could consider further simplification of some of these messages, but perhaps let's see how this goes.

[OCTRL-975](https://its.cern.ch/jira/browse/OCTRL-975)

### DIFF
--- a/core/task/task.go
+++ b/core/task/task.go
@@ -76,6 +76,7 @@ type parentRole interface {
 	ConsolidatedVarStack() (varStack map[string]string, err error)
 	CollectInboundChannels() []channel.Inbound
 	SendEvent(event.Event)
+	GetName() string
 }
 
 type Traits struct {
--- a/core/controlcommands/commandqueue.go
+++ b/core/controlcommands/commandqueue.go
@@ -101,7 +101,7 @@ func (m *CommandQueue) Start() {
 				}
 				if err == nil && response == nil {
 					log.WithField("partition", entry.cmd.GetEnvironmentId().String()).
-						Error("nil response")
+						Errorf("did not receive neither response nor error for %s", entry.cmd.GetName())
 				}
 
 				entry.callback <- response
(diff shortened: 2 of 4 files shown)


core/task/task.go
```diff
@@ -76,6 +76,7 @@ type parentRole interface {
 	ConsolidatedVarStack() (varStack map[string]string, err error)
 	CollectInboundChannels() []channel.Inbound
 	SendEvent(event.Event)
+	GetName() string
```
justonedev1, line 79: I would like to have some clarification. Is this implemented anywhere? Because I see adding new method to the interface but not implementing it.  Is it just done so you can access name from parentRole because inheriting classes implement it either way?

knopers8: Well, technically golang interfaces are not "inheritance", but yes, this method is already implemented by types which attempt to provide `parentRole` interface.

justonedev1: yeah you are right, "implements" is the proper term here. Thanks.

justonedev1 (approved):

teo (approved):

PR #655, opened 2025-02-25
[OCTRL-985] Place kafka messages related to same environments in the same partition

justonedev1:
- added key to messages containing environment id or run number
- changed Balancer from LeastBytes to Hash which makes kafka-go puts messages with same key to the same partition, fixing ordering problems

I did not add any tests, because I was unsure how to test this properly, so I just tested this on my openstack machine.

### DIFF
--- a/common/event/writer.go
+++ b/common/event/writer.go
@@ -62,7 +62,7 @@ func NewWriterWithTopic(topic topic.Topic) *KafkaWriter {
 		Writer: &kafka.Writer{
 			Addr:                   kafka.TCP(viper.GetStringSlice("kafkaEndpoints")...),
 			Topic:                  string(topic),
-			Balancer:               &kafka.LeastBytes{},
+			Balancer:               &kafka.Hash{},
 			AllowAutoTopicCreation: true,
 		},
 	}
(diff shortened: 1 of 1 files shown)


common/event/writer.go
```diff
@@ -111,36 +125,44 @@ func (w *KafkaWriter) WriteEventWithTimestamp(e interface{}, timestamp time.Time
...
+			key = extractAndConvertEnvID(e)
 			wrappedEvent = &pb.Event{
 				Timestamp:     timestamp.UnixMilli(),
 				TimestampNano: timestamp.UnixNano(),
 				Payload:       &pb.Event_IntegratedServiceEvent{IntegratedServiceEvent: e},
 			}
 		case *pb.Ev_RunEvent:
+			// we are using RunNumber as a key here, requested by M.Boulais
+			key = make([]byte, 4)
+			binary.BigEndian.PutUint32(key, e.RunNumber)
```
knopers8, line 165: Do i correctly understand that the consequence of this is that we will not have synchronization between `RunEvents` and all the others, as they use different keys? What is the advantage of using run number here, if env ID should be anyway available?

justonedev1: that is the reason for the comment. Martin wanted to use run number for this event in ticket, so I used it

justonedev1: by the way, RunEvents are in their own topic

knopers8: as discussed in person, please cross-check with Martin what is the reason for this exception just to be sure, but I'm OK with this.

justonedev1: As discussed in person: Martin is on vacation this week, I sent him a message. I will report when he is back.

knopers8: we are not releasing this week, so it's anyway fine to wait.

justonedev1: Martin agreed that using envId is fine here. So I did amend previous commit and reworded it.

knopers8: thank you!

knopers8 (approved):

PR #656, opened 2025-02-27
OCTRL-980 Send enabled links to DCS for selected detectors

knopers8:
While there are unit tests for the logic to gather the links, communication with DCS remains untested, since I could not see a good way to do it without production. It seems DCS is not available on staging currently.

This being said, the old way of setting active links is preserved, so we can optionally enable the feature in some organized test.


justonedev1 (approved):
I am approving these with the asterisk that we need to check the DCS communication.

knopers8:
Yes, absolutely. We will need to organize a test with Filippo.

PR #658, opened 2025-02-28
[core] ODC: Send SOR/EOR timestamps as FMQ properties

knopers8:
OCTRL-987

This involved extracting `setProperties` as a separate helper to share it between `Start` and `Stop`, and adding it to `Stop`, which was not the case. Contrarily to `Start`, we do not abandon `odc.Stop` in case `setProperties` fails, so we still have an attempt to stop the run.

Tested on staging, it looks like `setProperties` takes 50ms when using 2 EPNs. Looking at the past logs in production, it usually takes no more than 100ms, so I don't think anyone should complain, while this lets us use correct EOR time in QC.

### DIFF
--- a/core/integration/odc/plugin.go
+++ b/core/integration/odc/plugin.go
@@ -1363,6 +1363,12 @@ func (p *Plugin) CallStack(data interface{}) (stack map[string]interface{}) {
 				WithField("call", "Start").
 				Warn("cannot acquire FairMQ devices cleanup count for ODC")
 		}
+		runStartTimeMs, ok := varStack["run_start_time_ms"]
+		if !ok {
+			log.WithField("partition", envId).
+				WithField("call", "Start").
+				Warn("cannot acquire run_start_time_ms")
+		}
 
 		var (
 			runNumberu64 uint64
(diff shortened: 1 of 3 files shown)


core/integration/odc/plugin.go
```diff
@@ -1409,32 +1416,40 @@ func (p *Plugin) CallStack(data interface{}) (stack map[string]interface{}) {
...
 		}
+		runEndTimeMs, ok := varStack["run_end_time_ms"]
+		if !ok {
+			log.WithField("partition", envId).
+				WithField("call", "Start").
+				Warn("cannot acquire run_end_time_ms")
+		}
+
+		arguments := make(map[string]string)
+		arguments["run_end_time_ms"] = runEndTimeMs
```
justonedev1, line 1446: I have a question about this. Is this documented parameter that just wasn't used before, or are you introducing something new?

justonedev1 (commented):
I have one question, but code itself looks legitimate

core/integration/odc/plugin.go
knopers8: I'm not sure I understand the question. This parameter was not sent before to ODC, thus it could not be used, so it is something new.

justonedev1: Sorry for not being specific enough. I know that it wasn't used before. I am asking whether `run_end_time_ms` is existing parameter of ODC or something totally new, that needs to be implemented in ODC

knopers8: Ah no, these parameters are transparent to ODC, it just receives them and propagates to the tasks on EPNs. It's just key-values that tasks may use if they want to, e.g. QC is going to use it if it's available.

justonedev1: okay, thanks

justonedev1 (approved):

PR #668, opened 2025-03-18
Support newer versions

ktf:


ktf:
We need this to be able to bump to the new protobuf and gRPC (which is needed to be able to bump to the new ONNXRuntime).

ktf:
@davidrohr as we discussed over coffee.

ktf:
@knopers8 can we get a new tag with this?

justonedev1:
Was this tested?

ktf:
@justonedev1 The only tests I know about are the CI. Of course feel free to do any extra validation before tagging.

knopers8:
This bit is not tested in the CI, one will have to validate this manually.

knopers8 (approved):
Thanks. Given that the RPM builds are not working at the moment, the best I can do is to create a patch release and test it with the FLP suite CI.

PR #672, opened 2025-03-20
[core] More concise logging calls in DCS plugin

knopers8:


justonedev1 (approved):
Looks good, I was also trying to find some global solution how to make more concise logging calls, but I wasn't able to figure out something much smarter than what you did here.

PR #677, opened 2025-03-27
[core] fix a recursive mutex locking in task

knopers8:
Fixes a bug reported in OCTRL-999.

The recursive locking happens in the following trace:
```
 5  0x000000000122c829 in github.com/AliceO2Group/Control/core/task.(*Task).GetParent
    at /home/alibuild/sw/BUILD/eb488922a0fabb8512ec137d3a765b22d2941067/Control-Core/go/src/github.com/AliceO2Group/Control/core/task/task.go:741
 6  0x000000000122597c in github.com/AliceO2Group/Control/core/task.(*Task).GetTraits
    at /home/alibuild/sw/BUILD/eb488922a0fabb8512ec137d3a765b22d2941067/Control-Core/go/src/github.com/AliceO2Group/Control/core/task/task.go:248
 7  0x000000000122a217 in github.com/AliceO2Group/Control/core/task.(*Task).SendEvent
    at /home/alibuild/sw/BUILD/eb488922a0fabb8512ec137d3a765b22d2941067/Control-Core/go/src/github.com/AliceO2Group/Control/core/task/task.go:528
    ```

### DIFF
--- a/core/task/task.go
+++ b/core/task/task.go
@@ -251,6 +251,16 @@ func (t *Task) GetTraits() Traits {
 	return Traits{}
 }
 
+// lock-free version of GetTraits
+func (t *Task) getTraits() Traits {
+	if class := t.GetTaskClass(); class != nil {
+		if t.parent != nil {
+			return t.parent.GetTaskTraits()
+		}
+	}
+	return Traits{}
+}
+
 // Returns a consolidated CommandInfo for this Task, based on Roles tree and
 // Class.
 func (t *Task) BuildTaskCommand(role parentRole) (err error) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: core/task/task.go [hunk 1]
```diff
@@ -251,6 +251,16 @@ func (t *Task) GetTraits() Traits {
...
+// lock-free version of GetTraits
+func (t *Task) getTraits() Traits {
+	if class := t.GetTaskClass(); class != nil {
+		if t.parent != nil {
+			return t.parent.GetTaskTraits()
+		}
+	}
+	return Traits{}
+}
+
```
justonedev1, line 263: I don't like how it is literally the same code for `GetTraits()` and `getTraits()` except `t.parent` part. I'd probably  make `getTraits(*parentRole)` and call it from `GetTraits`, because the `t.GetParent()` usage in this case is not thread safe either way. We call it twice, so if there is a change of a parent from other goroutine, we might get different results on both calls due to race condition. With this change we will at least alway address the same parent, even if it changes half during the `GetTraits`  call.

justonedev1 (changes requested):

core/task/task.go
knopers8: did you mean something like this?

justonedev1 (approved):

PR #678, opened 2025-04-01
[OCTRL-948] Failing to deploy non-critical tasks should not prevent deployment

justonedev1:
INVARIANT status is introduced to fix OCTRL-948. INVARIANT is overwritten by any other status in status product (X() function). We are taking account the criticality of the role when merging. If all roles are noncritical we are returning INVARIANT. If there is any critical role(s) we are merging and returning their status, which shouldn't be INVARIANT. All roles except taskRole and callrole are taken as critical


justonedev1:
I addressed the comments and also did the tests on staging. But if you aprove this, dont merge I need to squash the fixup commit

knopers8 (approved):
thank you!

PR #680, opened 2025-04-02
[OCTRL-998]: Sending kafka messages in wrong order

justonedev1:
There was a goroutine in `WriteEventWithTimestamp` which can cause race condition resulting in problems mentioned in ticket OCTRL-998. 

However, there can be performance problems caused by removing this concurrence, as every Kafka write call will wait until the message is send by the writer. This can be prevented by setting writer into async mode. But doing so we won't get any errors that might happen while sending messages. From kafka-go/writer.go Async comment:
```
	// Setting this flag to true causes the WriteMessages method to never block.
	// It also means that errors are ignored since the caller will not receive
	// the returned value. Use this only if you don't care about guarantees of
	// whether the messages were written to kafka.
```

### DIFF
--- a/common/event/writer.go
+++ b/common/event/writer.go
@@ -97,91 +97,89 @@ func (w *KafkaWriter) WriteEventWithTimestamp(e interface{}, timestamp time.Time
 		return
 	}
 
-	go func() {
-		var (
-			err          error
-			wrappedEvent *pb.Event
-			key          []byte = nil
-		)
+	var (
+		err          error
+		wrappedEvent *pb.Event
+		key          []byte = nil
+	)
 
-		switch e := e.(type) {
-		case *pb.Ev_MetaEvent_CoreStart:
-			wrappedEvent = &pb.Event{
-				Timestamp:     timestamp.UnixMilli(),
-				TimestampNano: timestamp.UnixNano(),
-				Payload:       &pb.Event_CoreStartEvent{CoreStartEvent: e},
-			}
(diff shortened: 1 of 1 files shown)


knopers8 (changes requested):
Indeed I am afraid this will have a performance hit and might block the core in case of any event writer hiccup. Could we use a buffered channel of a reasonable length (100-1000?) and do non-blocking sends in `WriteEvent` and `WriteEventWithTimestamp`, while there would be a goroutine on the other side to consume and perform writing?

justonedev1:
Good idea, will do

justonedev1:
I did it... but I guess that it is the same functionality as just Saying Async = true for the kafka-go library... except we can print errors if we encounter some during writing... it would be possible to create some error checker (eg. store encountered errors in kafkaWriter and check if he encountered some periodically), but I don't know if it is realy necessary and if error message in IL is enough, especially when there will be a lot of them if we cannot write into kafka

common/event/writer.go
```diff
@@ -202,9 +216,7 @@ func (w *KafkaWriter) doWriteEvent(key []byte, e *pb.Event) error {
 		message.Key = key
 	}
 
-	err = w.WriteMessages(context.Background(), message)
-	if err != nil {
-		return fmt.Errorf("failed to write event: %w", err)
-	}
+	w.toWriteChan <- message
```
knopers8, line 219: This will block if the buffer is full. Should we make it non-blocking perhaps?

knopers8 (approved):

PR #681, opened 2025-04-02
Refactor kafka write

justonedev1:
This is based on OCTRL-998 branch, so please merge that first. Otherwise it should be just refactoring of Write method in kafka writer, so we don't repeat Timestamp manipulation everywhere, but only once, resulting in more clear code.


justonedev1:
closing this as it will be part of a bigger refactor of kafka writer

PR #701, opened 2025-05-05
[odc] plumbing for expendable and rmsjobid in odcdevice

justonedev1:


knopers8 (changes requested):
coudl you also add it to `deviceStateChangedEventPayload` and corresponding functions which fill it?

knopers8 (approved):

PR #705, opened 2025-05-08
[OCTRL-1016]:  Missing runNumber in KafkaEvent on topic aliecs.run

justonedev1:
I moved resetting of currentRunNumber from `StartActivityTransition` object if some tasks fail to start to `GO_ERROR` transition after message of EOEOR is send.

@knopers8 Do you think that this can break something? Or do you propose something else?

### DIFF
--- a/core/environment/environment.go
+++ b/core/environment/environment.go
@@ -614,14 +614,6 @@ func newEnvironment(userVars map[string]string, newId uid.ID) (env *Environment,
 					errorMsg = e.Err.Error()
 				}
 
-				if e.Event == "STOP_ACTIVITY" {
-					// If the event is STOP_ACTIVITY, we remove the active run number after all hooks are done.
-					env.workflow.GetVars().Set("last_run_number", strconv.Itoa(int(env.currentRunNumber)))
-					env.currentRunNumber = 0
-					env.workflow.GetVars().Del("run_number")
-					env.workflow.GetVars().Del("runNumber")
-				}
-
 				// publish transition step complete event
 				the.EventWriterWithTopic(topic.Environment).WriteEvent(&pb.Ev_EnvironmentEvent{
 					EnvironmentId:        env.id.String(),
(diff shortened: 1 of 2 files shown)


core/environment/environment.go
```diff
@@ -602,6 +602,7 @@ func newEnvironment(userVars map[string]string, newId uid.ID) (env *Environment,
 							Debug("O2 End Completion time already set before after_GO_ERROR")
 					}
 					env.invalidateAutoStopTransition()
+					env.currentRunNumber = 0
```
knopers8, line 605: 20 lines lower there is this:
```
				if e.Event == "STOP_ACTIVITY" {
					// If the event is STOP_ACTIVITY, we remove the active run number after all hooks are done.
					env.workflow.GetVars().Set("last_run_number", strconv.Itoa(int(env.currentRunNumber)))
					env.currentRunNumber = 0
					env.workflow.GetVars().Del("run_number")
					env.workflow.GetVars().Del("runNumber")
				}
	```
Shouldn't we just modify it so that it also deletes run numbers in case of GO_ERROR? We have to keep in mind though that GO_ERROR may not necessarily happen during RUNNING, but also in other states.

Comment (knopers8) [reply core/environment/environment.go:605]: I am suggesting this, because the GO_ERROR hooks with positive weights should still have access to run number.

File: core/environment/environment.go [hunk 2]
```diff
@@ -614,8 +614,8 @@ func newEnvironment(userVars map[string]string, newId uid.ID) (env *Environment,
 					errorMsg = e.Err.Error()
 				}
 
-				if e.Event == "STOP_ACTIVITY" {
-					// If the event is STOP_ACTIVITY, we remove the active run number after all hooks are done.
+				if e.Event == "STOP_ACTIVITY" || e.Event == "ERROR" {
```
knopers8, line 617: is the event ERROR or GO_ERROR?

justonedev1: GO_ERROR I made a mistake and forgot to push correct change while writing comment for you

justonedev1:
We talked about multiple possibilities, but I kinda don't understand why just adding `GO_ERROR` isn't enough here, it should happen after all hooks are handled.

core/environment/environment.go
```diff
@@ -614,8 +614,8 @@ func newEnvironment(userVars map[string]string, newId uid.ID) (env *Environment,
 					errorMsg = e.Err.Error()
 				}
 
-				if e.Event == "STOP_ACTIVITY" {
-					// If the event is STOP_ACTIVITY, we remove the active run number after all hooks are done.
+				if e.Event == "STOP_ACTIVITY" || e.Event == "GO_ERROR" {
+					// If the event is STOP_ACTIVITY or GO_ERROR, we remove the active run number after all hooks are done.
 					env.workflow.GetVars().Set("last_run_number", strconv.Itoa(int(env.currentRunNumber)))
```
knopers8, line 619: i think we should avoid setting `"last_run_number"` if `env.currentRunNumber` is 0. This could happen if we e.g. reach ERROR without ever starting a run.

This way, it will be less risky to have a bug due to some piece of code assuming that non-empty `last_run_number` means we had a run before.

knopers8 (approved):

PR #709, opened 2025-05-19
[core] split TRG cleanup into two calls for distinguishable kafka events

knopers8:
This commit splits trg.Cleanup() into two calls which should be used subsequently instead of trg.Cleanup. Thanks to this, the GUIs can listen to kafka events for a specific operationName ("trg.EnsureRunStop") for trigger stop timestamp. We preferred to avoid having the GUIs hunt for operationStep "perform TRG call: RunStop", which seems to be a message targeted for humans and could be more prone to change, thus breaking behaviour in GUIs.

Fixes OCTRL-1017.

### DIFF
--- a/core/integration/trg/plugin.go
+++ b/core/integration/trg/plugin.go
@@ -1285,6 +1285,7 @@ func (p *Plugin) CallStack(data interface{}) (stack map[string]interface{}) {
 		return runUnloadFunc(ctx, runNumber64)
 	}
 	stack["Cleanup"] = func() (out string) {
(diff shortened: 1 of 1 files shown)


justonedev1 (commented):
You are commenting that you are splitting TRG cleanup, but I don't see something being split, just that you are adding new two calls. Is it on purpose or do I just misunderstand?

justonedev1:
Ah, okay. I overlooked the obsolete comment, sorry

justonedev1 (approved):

PR #710, opened 2025-05-19
[OCTRL-1018] Once NewEnvironmentAsync returns, the env should exist

justonedev1:


justonedev1:
I split `CreateEnvironment` to Create and Run, so we can change behaviour in `NewEnvironmentAsync` according to George's desires.

core/server.go
```diff
@@ -436,6 +455,20 @@ func (m *RpcServer) NewEnvironment(cxt context.Context, request *pb.NewEnvironme
...
 	}
 
+	err = m.state.environments.RunEnvironment(request.GetWorkflowTemplate(), id, request.GetAutoTransition())
+	if err != nil {
+		st := status.Newf(codes.Internal, "cannot run new environment: %s", utils.TruncateString(err.Error(), MAX_ERROR_LENGTH))
+		ei := &pb.EnvironmentInfo{
+			Id:           id.String(),
+			CreatedWhen:  time.Now().UnixMilli(),
+			State:        "ERROR", // not really, but close
+			NumberOfFlps: 0,
```
knopers8, line 465: would it be difficult to have a correct value for the `NumberOfFlps`?

core/server.go
```diff
@@ -374,7 +374,26 @@ func (m *RpcServer) NewEnvironmentAsync(cxt context.Context, request *pb.NewEnvi
 	}
 	defer setCurrentUnixMilli(&reply.Timestamp)
 
-	go m.doNewEnvironmentAsync(cxt, userVars, request.GetRequestUser(), request.GetWorkflowTemplate(), request.GetPublic(), request.GetAutoTransition(), id)
+	public := request.GetPublic()
+	workflowTemplate := request.GetWorkflowTemplate()
+	requestUser := request.GetRequestUser()
+
+	id, err = m.state.environments.CreateEnvironment(request.GetWorkflowTemplate(), userVars, request.GetPublic(), id)
```
knopers8, line 381: One has to recognize that this will move environment creation into the synchronous part, which is a change in the behaviour. I imagine that the creation of an environment with all FLPs and no cached workflow templates may take 1-2 minutes. Please make sure that this is OK with George.

justonedev1: I know, but this is the only way how to achieve George's desired behavior without changing order of operations and other code changes. 

And it is also the reason why I gave binary with these changes to STG for him to play with so he can see how it will behave.

core/server.go
justonedev1: no idea, I didn't change anything, I just split the code.

justonedev1:
I am closing this as George doesn't want this if it takes too long and he will add some workaround in GUI

PR #712, opened 2025-05-22
[core] update odc.proto to include collection info

knopers8:
OCTRL-1024


knopers8:
WIP until ODC v0.84.3 is used at P2.

justonedev1 (approved):
looks ok

PR #720, opened 2025-06-03
[OCTRL-1012]: GetEnvironments is missing ODC devices info even with flag set

justonedev1:


core/protos/o2control.proto
```diff
@@ -123,6 +123,7 @@ message TeardownReply {}
 message GetEnvironmentsRequest {
     bool showAll = 1;
     bool showTaskInfos = 2;
+    bool showIntegratedServices = 3;
```
knopers8, line 126: I think the flag name is unclear. Even when it's false, we fill `ingratedServicesData` with the output of `GetEnvironmentsShortData`, while the flag name gives impression we will get nothing. Could we name it for example `showDetailedIntegratedServices`, or anything conveying this meaning? A short comment might also help.

The fact that `showAll` does not imply we show task infos is another issue, but it's something that was there before.

core/server.go
```diff
@@ -294,6 +294,11 @@ func (m *RpcServer) GetEnvironments(cxt context.Context, request *pb.GetEnvironm
 		if request.GetShowTaskInfos() {
 			e.Tasks = tasksToShortTaskInfos(tasks, m.state.taskman)
 		}
+
+		if request.GetShowIntegratedServices() {
+			e.IntegratedServicesData = integration.PluginsInstance().GetEnvironmentsData([]uid.ID{env.Id()})[env.Id()]
+		}
+
```
knopers8, line 301: Is there a reason to put this code here instead of combining it with this call from above?
```
integratedServicesEnvsData := integration.PluginsInstance().GetEnvironmentsShortData(m.state.environments.Ids())
```
Depending on the flag, we could call either `GetEnvironmentsShortData` or `GetEnvironmentsData` instead of calling always the first, and optionally overwriting it with the 2nd.

PR #725, opened 2025-06-13
Avoid specifying the FairMQ version

ktf:
It does not work and it's in any case obsolete.


ktf:
@knopers8 @justonedev1 could you please merge and tag?

PR #731, opened 2025-08-05
No need to force ERROR when GO_ERROR succeeded

knopers8:
Probably due to a mostly harmless mistake, we would force an environment into ERROR state even when not needed, after a successful GO_ERROR. If very unlucky, this could cause some conflicts with following transitions towards environment destruction.

This affects only GO_ERROR due to ODC going to ERROR, other cases are fine.

I did not see this bug cause any troubles, I just accidentally stumbled upon it.


justonedev1 (approved):
nice catch

PR #732, opened 2025-08-06
Call ODC STOP only when the partition is in RUNNING

knopers8:
This allows us to stop a run on EPNs at GO_ERROR transition by adding a corresponding ODC.EnsureStop hook.
As GO_ERROR can occur with any source state, we make sure to make the actual STOP call only if the ODC partition is in RUNNING.
    
At the same time, ODC partitions require us to call ODC.Stop if they voluntarily transition to ERROR.
In such case, ODC.Stop allows the remaining healthy devices to finish processing.
By keeping the original ODC.Stop behaviour, we preserve this functionality.
    
Additionally, the commit includes minor corrections to a few related logs.
    
Fixes OCTRL-1036.

### DIFF
--- a/core/integration/odc/plugin.go
+++ b/core/integration/odc/plugin.go
@@ -1437,12 +1437,13 @@ func (p *Plugin) CallStack(data interface{}) (stack map[string]interface{}) {
 		rn, ok := varStack["run_number"]
 		if !ok {
 			log.WithField("partition", envId).
-				WithField("call", "Start").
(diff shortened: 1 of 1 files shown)


justonedev1 (approved):
So apart from adding the stop call to hooks it was necessary to change the plugin code as well... well done!

knopers8:
Putting on hold until we get confirmation whether we should or should not call ODC STOP when a partition goes to ERROR.

PR #747, opened 2025-08-21
Fix race condition in tests

justonedev1:
This patch fixes the `panic: runtime error: invalid memory address or nil pointer dereference` error triggering in tests. It was caused by the racecondition between `Run()` and `Stop()`. `Run` is blocking so it has to be called in goroutine (that is on purpose so we can handle errors in core) so sometimes `Stop()` is called to quickly before `Run` is done, so you could call `ListenAndServe` on an empty object. 
Usage of atomic pointer is not strictly necessary for the fix of mentioned problem. But it assures that no goroutine will be stuck if we change global server too quickly. Usage of the atomic pointer doesn't cause any performance penalty as it just holds the global server and nothing more.

This bug couldn't influence production as we didn't call `Run()` and immediately `Stop()`.

### DIFF
--- a/common/monitoring/monitoring.go
+++ b/common/monitoring/monitoring.go
@@ -155,26 +155,26 @@ func handleFunc(endpointName string) {
 //
 // If we attempt send more messages than the size of the buffer, these overflowing messages will be ignored and warning will be logged.
 func Run(port uint16, endpointName string) error {
-	srv := &http.Server{Addr: fmt.Sprintf(":%d", port)}
+	localServer := &http.Server{Addr: fmt.Sprintf(":%d", port)}
 	// only one Run should initialize and serve
-	if !server.CompareAndSwap(nil, srv) {
+	if !server.CompareAndSwap(nil, localServer) {
 		return nil
 	}
 	initChannels()
 	go eventLoop()
 	handleFunc(endpointName)
 	// block until Shutdown is called
-	return srv.ListenAndServe()
+	return localServer.ListenAndServe()
 }
 
 func Stop() {
-	srv := server.Swap(nil)
-	if srv == nil {
+	localServer := server.Swap(nil)
+	if localServer == nil {
 		return
 	}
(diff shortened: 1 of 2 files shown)


common/monitoring/monitoring.go
```diff
@@ -154,33 +155,30 @@ func handleFunc(endpointName string) {
 //
 // If we attempt send more messages than the size of the buffer, these overflowing messages will be ignored and warning will be logged.
 func Run(port uint16, endpointName string) error {
-	if IsRunning() {
+	srv := &http.Server{Addr: fmt.Sprintf(":%d", port)}
+	// only one Run should initialize and serve
+	if !server.CompareAndSwap(nil, srv) {
```
knopers8, line 160: Sorry, I get very much confused when trying to read this and remembering which server/srv is which. Could you perhaps call them `serverAtomicPtr` and just `server`, or anything else which conveys the difference?

common/monitoring/monitoring.go
```diff
@@ -154,33 +155,30 @@ func handleFunc(endpointName string) {
...
-
 	initChannels()
-
 	go eventLoop()
-
-	server = &http.Server{Addr: fmt.Sprintf(":%d", port)}
 	handleFunc(endpointName)
-	return server.ListenAndServe()
+	// block until Shutdown is called
+	return srv.ListenAndServe()
```
knopers8, line 167: Will the test exit with `Stop` gets a `nil` and returns before we reach this? What will call `srv.Shutdown()` in such case?

justonedev1: as talked about personally, this doesn't matter because we run all `Run()` as separate goroutines, so the worst thing that might happen is leaked goroutine which will be cleaned by the OS. I thought that I would fix it with using mutex to guard `server`, but as `Run` is blocking I would have to rework the whole functionality of `Run()` and `Stop()`. This is probably excessive effort just to fix leak that will not even happen in current implementation of tests, where we are always waiting until the monitoring is up and running.

knopers8 (approved):

PR #751, opened 2025-09-18
OCTRL-678 Pass the error message to BKP in case of GO_ERROR

knopers8:
An error message is now propagated to kafka events just before scheduling a GO_ERROR transition.

### DIFF
--- a/core/environment/environment.go
+++ b/core/environment/environment.go
@@ -1176,6 +1176,15 @@ func (env *Environment) QueryRoles(pathSpec string) (rs []workflow.Role) {
 	return
 }
 
+func (env *Environment) GetId() uid.ID {
+	if env == nil {
+		return ""
+	}
+	env.Mu.RLock()
+	defer env.Mu.RUnlock()
+	return env.id
+}
+
 func (env *Environment) GetPath() string {
 	return ""
 }
@@ -1222,7 +1231,12 @@ func (env *Environment) subscribeToWfState(taskman *task.Manager) {
 								log.WithField("partition", env.id).
 									WithField("level", infologger.IL_Ops).
 									Error("one of the critical tasks went into ERROR state, transitioning the environment into ERROR")
+
+								the.EventWriterWithTopic(topic.Environment).WriteEvent(
+									NewEnvGoErrorEvent(env, newCriticalTasksErrorMessage(env)),
+								)
 								err := env.TryTransition(NewGoErrorTransition(taskman))
+
 								if err != nil {
 									if env.Sm.Current() == "ERROR" {
 										log.WithField("partition", env.id).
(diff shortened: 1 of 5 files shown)


core/environment/environment.go
```diff
@@ -1176,6 +1176,15 @@ func (env *Environment) QueryRoles(pathSpec string) (rs []workflow.Role) {
 	return
 }
 
+func (env *Environment) GetId() uid.ID {
+	if env == nil {
+		return ""
+	}
+	env.Mu.RLock()
```
justonedev1, line 1183: Are you sure that you cannot cause some deadlocks by this lock acquiring? I remember that there were some problems in the past. (same for the `Task.GetState()`

core/environment/utils.go
```diff
@@ -101,3 +105,37 @@ func sortMapToString(m map[string]string) string {
...
+}
+
+func newCriticalTasksErrorMessage(env *Environment) string {
+	criticalTasksInError := env.workflow.GetTasks().Filtered(func(t *task.Task) bool {
+		return t.GetTraits().Critical && t.GetState() == sm.ERROR
+	})
+
+	if len(criticalTasksInError) == 0 {
+		return "no critical tasks in ERROR"
+	} else if len(criticalTasksInError) == 1 {
```
justonedev1, line 128: why is one failed task so significant that it requires specific case?

justonedev1 (commented):
Is there a reason why you are not sending the error message directly from GO_ERROR transition from the environment?

core/environment/utils.go
knopers8: Because it's the usual case and it allows us to easily say which task crashed and where. With multiple tasks it becomes more difficult (i.e. you cannot easily know which one was first, the message might become very long if many tasks are in ERROR), so I prefered not to overcomplicate it.

core/environment/environment.go
knopers8: I am almost certain it should not cause trouble here. The convention which I think works is: exported methods should use locks, internal methods should not. If an internal one needs to call an exported one, one should instead create an internal version of a method (e.g. `getId()`). This way, we ensure that there are no recursive lock attempts and any external calls are properly mutexed.

knopers8:
> Is there a reason why you are not sending the error message directly from GO_ERROR transition from the environment?

Yes, for two reasons:
 - it's not known anymore, but we could work it around by adding it as a string argument in `NewGoErrorTransition` and adding the call to `GoErrorTransition.do()`. But...
 - If a critical hook for `before_GO_ERROR` fails, we will never reach `GoErrorTransition.do()`.

core/environment/environment.go
justonedev1: alright, honestly didn't check whether it can do a problem here. I just remember that we had problems with Lock RLocked mutex (or vice versa) and I just wanted to ask.

justonedev1 (approved):

PR #752, opened 2025-10-07
[O2B-1475] Add proto message and associated ones for BeamModeEvent changes

graduta:
PR which:
* adds a new event message in `events.proto` that is to be used by BKP-LHC-Client to send events related to STABLE_BEAMS changes which are to be consumed by ECS;
* adds a new message in `common.proto` for beam information that has currently been decided as needed by ECS


common/protos/events.proto
```diff
@@ -127,6 +127,15 @@ message Ev_RunEvent {
   common.User lastRequestUser = 8;
 }
 
+/**
+ * Beam mode changes are propagated as Kafka events and to be sent by the BKP-LHC-Client on a dedicated topic
+ * e.g. dip.lhc.beam_mode
+ */
+message Ev_BeamModeEvent {
```
graduta, line 134: This name `Ev_BeamModeEvent` follows the naming convention for the test of events.

I personally, consider that the `Ev` already stands for Event and I would recommend something like `Ev_BeamModeChange`

What do you @knopers8 and @justonedev1 think?

justonedev1:
Just a note: This PR needs to generate go representation of protofiles and @graduta asked us to do so, after we agree on this PR.

common/protos/events.proto
justonedev1: I think that you are right, but I would keep the same convention as the rest of ECS. So I'd keep it as it is now.

PR #798, opened 2026-02-19
Toy model for imitating O2 software memory usage in K8s

knopers8:
This directory contains a toy model which pretends to be O2 software in terms of how it uses memory and runs on a K8s cluster. Closes OCTRL-1079, see also more info there.


knopers8:
The scripts were mostly vibe-coded and do what I wanted them to do, but please don't feel obliged to review them in detail, they are surely not production-grade.

justonedev1 (approved):
Thanks for the demonstration, it is good work

PR #800, opened 2026-02-27
[OCTRL-1082] fixed race conditions caused by repeated use of GetParent

justonedev1:


core/task/task.go
```diff
@@ -718,13 +720,14 @@ func (t *Task) GetMesosCommandTarget() controlcommands.MesosCommandTarget {
...
-	defer t.mu.RUnlock()
-
 	if t == nil {
 		log.Warn("attempted to get properties of nil task")
 		return make(map[string]string)
 	}
+
+	t.mu.RLock()
+	defer t.mu.RUnlock()
+
```
knopers8, line 730: nice catch

knopers8 (approved):
thank you!

PR #801, opened 2026-03-02
Fixes for boost 1.90

davidrohr:
This makes it compile with boost 1.90 for now. But we should modernize the boost usage in Control! Eventually, some of the deprecated boost features like the boost random number generators will be removed, we should switch to std::...
And the boost lockfree queue is used incorrectly, it must not be called with a trivial constructor, if no compile-time capacity is provided.


knopers8 (approved):
Thank you! I will note the wrong use of the queue.

PR #803, opened 2026-03-15
[peanut] peanut now supports direct, fmq batched and single step modes

justonedev1:
during my quest for kubernetes I had a requirement for tool that can control fmq processes in all modes that exist inside ecs:
- direct occ mode
- fairmq mode: same as ecs where we batch some commands to one ecs transition
- fairmq step mode: granular fairmq state machine control

peanut now provides TUI and CLI modes for all described modes
added reconnecting

changes done by Claude Sonnet 4.6 without my manual intervention, I just tested if all changes work. I don't deem this tool important enough to understand properly the code.

### DIFF
--- a/cmd/peanut/main.go
+++ b/cmd/peanut/main.go
@@ -40,16 +40,60 @@ func main() {
 		fmt.Fprint(os.Stderr, `peanut — process execution and control utility for OCC / FairMQ processes
 
 TUI mode (interactive, launched when no command is given):
-  OCC_CONTROL_PORT=<port> peanut
-  peanut -addr host:port -mode fmq
+  peanut                                   direct mode via OCC_CONTROL_PORT env var
+  peanut -addr host:port                   direct mode (OCC protobuf, one button per transition)
+  peanut -addr host:port -mode fmq         fmq batched mode (drives full FairMQ sequence per transition)
+  peanut -addr host:port -mode fmq-step    fmq single-step mode (one button per raw FairMQ event)
 
 CLI mode (non-interactive, launched when a command is given):
   peanut [flags] <command> [args]
-  Run "peanut -addr x get-state" for full CLI usage.
 
-Flags:
+TUI Flags:
+  -addr  string   gRPC address (host:port); if empty, uses OCC_CONTROL_PORT env var in direct mode
+  -mode  string   direct (default), fmq, or fmq-step
+
+CLI Flags:
+  -addr     string    gRPC address (default "localhost:47100")
+  -mode     string    fmq (default) or direct
+  -timeout  duration  unary call timeout (default 30s)
+  -config   string    path to YAML/JSON file with arguments to push (inline key=val args take precedence)
+
+CLI Commands:
+  get-state
+        Print the current FSM state.
+
+  transition <fromState> <toState> [key=val ...]
+        High-level OCC transition. In fmq mode drives the full multi-step FairMQ sequence:
+          STANDBY→CONFIGURED  runs INIT DEVICE, COMPLETE INIT, BIND, CONNECT, INIT TASK
+          CONFIGURED→RUNNING  runs RUN
+          RUNNING→CONFIGURED  runs STOP
+          CONFIGURED→STANDBY  runs RESET TASK, RESET DEVICE
+        In direct mode sends a single OCC protobuf Transition RPC.
+        key=val pairs are forwarded as ConfigEntry arguments.
+
+  direct-step <srcState> <event> [key=val ...]
+        Low-level: send a single raw OCC protobuf Transition RPC regardless of -mode.
+        Events: CONFIGURE, RESET, START, STOP, RECOVER, EXIT
+
+  fmq-step <srcFMQState> <fmqEvent> [key=val ...]
+        Low-level: send a single raw FairMQ gRPC Transition call regardless of -mode.
+        FairMQ state/event names that contain spaces must be quoted.
+
+  state-stream
+        Subscribe to StateStream; print updates until interrupted (ctrl-c to stop).
+
+  event-stream
+        Subscribe to EventStream; print events until interrupted (ctrl-c to stop).
+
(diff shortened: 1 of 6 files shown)


cmd/peanut/main.go
```diff
@@ -25,17 +25,46 @@
...
 
-	if err := peanut.Run(cmdString); err != nil {
-		panic(err)
+TUI mode (interactive, launched when no command is given):
+  OCC_CONTROL_PORT=<port> peanut
+  peanut -addr host:port -mode fmq
+
+CLI mode (non-interactive, launched when a command is given):
+  peanut [flags] <command> [args]
+  Run "peanut -addr x get-state" for full CLI usage.
```
knopers8: This does not list all commands and frankly I couldn't find a way to do it.

occ/peanut/README.md
```diff
@@ -1,42 +1,227 @@
...
+
+```bash
+peanut -addr localhost:47100 -mode direct direct-step STANDBY CONFIGURE key=val
+```
+
+Events: `CONFIGURE`, `RESET`, `START`, `STOP`, `RECOVER`, `EXIT`
+
+#### `fmq-step <srcFMQState> <fmqEvent> [key=val ...]`
+
+Low-level: send a single raw FairMQ gRPC Transition call (JSON codec).
```
knopers8, line 204: I don't quite understand how this is more low-level, if the only difference with high-level is that I have to type in an event instead of target state. Could you explain please?

knopers8 (changes requested):
Thanks, it indeed impressive that all of this was generated code. I have some comments for small interface improvements.

Also, when running a DPL task, I got the following crash for `state-stream`:
```
~/alice/Control$ bin/peanut -addr localhost:31004 state-stream
streaming state updates from localhost:31004  (ctrl-c to stop)
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x8dba6d]

goroutine 1 [running]:
github.com/AliceO2Group/Control/occ/peanut.RunCLI({0xc000020150, 0x3, 0x3})
	/home/pkonopka/alice/Control/occ/peanut/cli.go:162 +0x120d
main.main()
	/home/pkonopka/alice/Control/cmd/peanut/main.go:58 +0x1b4
```
That was the task:
```
OCC_CONTROL_PORT=31004 o2-testworkflows-diamond-workflow --id D_dds%TaskIndex%_%CollectionIndex% --shm-monitor false --log-color false --no-batch --color false --channel-config "name=from_B_to_D,type=pull,method=connect,address=ipc://@localhost189507_22003,transport=shmem,rateLogging=0,rcvBufSize=4,sndBufSize=4" --channel-config "name=from_C_to_D,type=pull,method=connect,address=ipc://@localhost189507_22004,transport=shmem,rateLogging=0,rcvBufSize=4,sndBufSize=4" --aBool --aDouble 3 --aFloat 2 --aString foobar --anInt 1 --aod-file "" --aod-memory-rate-limit 0 --aod-writer-json "" --aod-writer-keep "" --aod-writer-maxfilesize 0 --aod-writer-ntfmerge -1 --aod-writer-resdir "" --aod-writer-resfile "" --aod-writer-resmode RECREATE --bad-alloc-attempt-interval 50 --bad-alloc-max-attempts 1 --clone "" --early-forward-policy never --fairmq-ipc-prefix "@" --fairmq-rate-logging 0 --fairmq-recv-buffer-size 4 --fairmq-send-buffer-size 4 --forwarding-destination drop --forwarding-policy dangling --io-threads 1 --labels "" --pipeline "" --readers 1 --severity info --shm-allocation rbtree_best_fit --shm-metadata-msg-size 0 --shm-mlock-segment false --shm-mlock-segment-on-creation false --shm-no-cleanup false --shm-segment-id 0 --shm-segment-size 2000000000 --shm-throw-bad-alloc true --shm-zero-segment false --signposts "" --spawners 1 --stacktrace-on-signal simple --timeframes-rate-limit 0 --timeframes-rate-limit-ipcid -1 --workflow-suffix "" --session dpl_189507 --control-port 31004 -S $CONTROL_OCCPLUGIN_ROOT/lib/ -P OCClite
```

executor/executorcmd/nopb/occclient.go
```diff
@@ -61,8 +61,42 @@ func (c *occClient) EventStream(ctx context.Context, in *pb.EventStreamRequest,
...
+
+func (x *occStateStreamClient) Recv() (*pb.StateStreamReply, error) {
+	m := new(pb.StateStreamReply)
+	if err := x.ClientStream.RecvMsg(m); err != nil {
+		return nil, err
+	}
+	return m, nil
+}
+
 func (c *occClient) StateStream(ctx context.Context, in *pb.StateStreamRequest, opts ...grpc.CallOption) (pb.Occ_StateStreamClient, error) {
```
knopers8, line 76: Is this part used also for controlling tasks by the executor? Have you tried if this still works after the modification?

knopers8 (changes requested):
After the last commit, the TUI for FMQ mode stopped behaving 100% correctly, it used to work before (i confirmed by reverting by one commit).

In particular, when connecting with this:
```
OCC_CONTROL_PORT=31004 bin/peanut -addr localhost:31004 -mode fmq
```
I immediatelly see a window saying "Connection lost, Error:rpc error: code = Unimplemented desc =". I can still send commands and that works, but I am getting the aforementioned errors which I was not getting before.

executor/executorcmd/nopb/occclient.go
justonedev1: this is safe, because it isn't called anywhere outside the peanut. It just added the same logic as event stream. I didn't make sense that one was implemented and one wasn't...

knopers8: ok

knopers8 (approved):
thank you

PR #805, opened 2026-03-18
[OCTRL-1081] Wrap kubectl into Mesos executor task

justonedev1:
The task description is in the documentation `docs/kubernetes_ecs.md`

### DIFF
--- a/docs/kubernetes_ecs.md
+++ b/docs/kubernetes_ecs.md
@@ -23,6 +23,12 @@ is ability to run Pods with privileged permissions and also under user `flp`.
 This means that the machine has to have `flp` user setup the same way as
 if you would do the installation with [`o2-flp-setup`](https://alice-flp.docs.cern.ch/Operations/Experts/system-configuration/utils/o2-flp-setup/).
 
+## Task Controller
+
+Following text assumes that there is a Task Controller from `control-operator` running
+at your K8s cluster and Task CRD installed at your cluster.
+You can find the details about the usage in the [documentation](/control-operator/README.md).
+
 ## Running tasks (`KubectlTask`)
 
 ECS is setup to run tasks through Mesos on all required hosts baremetal with active
@@ -62,7 +68,7 @@ inside the `ControlWorkflows` directory. Namely:
 
 * add path to the kubectl manifest as the first argument in `.command.arguments` field
 * change `.control.mode` to either `kubectl_direct` or `kubectl_fairmq`
-You can find working template inside `control-operator/ecs-manifests/control-workflows/*_kube.yaml`
+You can find working template inside `control-operator/ecs-manifests/control-workflows/*-kube.yaml`
 
 Working kubectl manifests are to be found in `control-operator/ecs-manifests/kubernetes-manifests`.
 You can see `*test.yaml` for concrete deployable manifests by `kubectl apply`, the rest
--- a/docs/kubernetes_ecs.md
+++ b/docs/kubernetes_ecs.md
@@ -8,18 +8,18 @@
 While prototyping we used many Kubernetes clusters, namely [`kind`](https://kind.sigs.k8s.io/), [`minikube`](https://minikube.sigs.k8s.io/docs/) and [`k3s`](https://k3s.io/)
 in both local and remote cluster deployment. We used Openstack for remote deployment.
 Follow the guides at the individual distributions in order to create the desired cluster setup.
-For now we chose `k3s` for most of the activities performed because it is lightweight
+k3s is recommended to run this prototype, as it is lightweight
 and easily installed distribution which is also [`CNCF`](https://www.cncf.io/training/certification/) certified.
 
 All settings of `k3s` were used as default except one: locked-in-memory size. Use `ulimit -l` to learn
 what is the limit for the current user and `LimitMEMLOCK` inside the k3s systemd service config
 to set it for correct value. Right now the `flp` user has unlimited size (`LimitMEMLOCK=infinity`).
-This config is necessary because even if you are running PODs with the privileged security context
+This config is necessary because even if you are running Pods with the privileged security context
 under user flp, Kubernetes still sets limits according to its internal settings and doesn't
 respect linux settings.
 
 Another setup we expect at this moment to be present at the target nodes
-is ability to run PODs with privileged permissions and also under user `flp`.
+is ability to run Pods with privileged permissions and also under user `flp`.
 This means that the machine has to have `flp` user setup the same way as
 if you would do the installation with [`o2-flp-setup`](https://alice-flp.docs.cern.ch/Operations/Experts/system-configuration/utils/o2-flp-setup/).
 
@@ -30,7 +30,7 @@ task management (see [`ControllableTask`](/executor/executable/controllabletask.
 and OCC gRPC communication. When running docker task through ECS we could easily
 wrap command to be run into the docker container with proper settings
 ([see](/docs/running_docker.md)). This is however not possible for Kubernetes
-workloads as the PODs are "hidden" inside the cluster. So we plan
+workloads as the Pods are "hidden" inside the cluster. So we plan
 to deploy our own Task Controller which will connect to and guide
 OCC state machine of required tasks. Thus we need to create custom
 POC way to communicate with Kubernetes cluster from Mesos executor.
--- a/docs/kubernetes_ecs.md
+++ b/docs/kubernetes_ecs.md
@@ -1,0 +1,70 @@
+# ECS with Kubernetes
+
+> ⚠️ **Warning**
+> All Kubernetes work done is in a stage of prototype.
+
+## Kubernetes Cluster
+
+While prototyping we used many Kubernetes clusters, namely [`kind`](https://kind.sigs.k8s.io/), [`minikube`](https://minikube.sigs.k8s.io/docs/) and [`k3s`](https://k3s.io/)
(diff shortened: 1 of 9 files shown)


core/task/taskclass/class.go
```diff
@@ -154,13 +153,17 @@ func (c *Class) MarshalYAML() (interface{}, error) {
...
+	// 	aux.Control.Mode = "fairmq"
+	// } else if c.Control.Mode == controlmode.BASIC {
+	// 	aux.Control.Mode = "basic"
+	// } else if c.Control.Mode == controlmode.KUBECTL {
+	// 	aux.Control.Mode = "kubectl"
+	// } else {
+	// 	aux.Control.Mode = "direct"
+	// }
+
+	aux.Control.Mode = c.Control.Mode.String()
```
knopers8, line 155: There is a change in behaviour for hooks though, no? Before they were getting `direct` instead of `hook`, which actually smells like a bug, but perhaps something is relying on it?

docs/kubernetes_ecs.md
```diff
@@ -0,0 +1,70 @@
...
+
+> ⚠️ **Warning**
+> All Kubernetes work done is in a stage of prototype.
+
+## Kubernetes Cluster
+
+While prototyping we used many Kubernetes clusters, namely [`kind`](https://kind.sigs.k8s.io/), [`minikube`](https://minikube.sigs.k8s.io/docs/) and [`k3s`](https://k3s.io/)
+in both local and remote cluster deployment. We used Openstack for remote deployment.
+Follow the guides at the individual distributions in order to create the desired cluster setup.
+For now we chose `k3s` for most of the activities performed because it is lightweight
```
knopers8: That might give an impression that there was some kind of decision process in ALICE or FLP. Perhaps rephrase to:

> `k3s` is recommended to run this prototype, as it is a lightweight and...

docs/kubernetes_ecs.md
```diff
@@ -0,0 +1,70 @@
...
+and easily installed distribution which is also [`CNCF`](https://www.cncf.io/training/certification/) certified.
+
+All settings of `k3s` were used as default except one: locked-in-memory size. Use `ulimit -l` to learn
+what is the limit for the current user and `LimitMEMLOCK` inside the k3s systemd service config
+to set it for correct value. Right now the `flp` user has unlimited size (`LimitMEMLOCK=infinity`).
+This config is necessary because even if you are running PODs with the privileged security context
```
knopers8:
```suggestion
This config is necessary because even if you are running Pods with the privileged security context
```
It's a detail, but "Pod" and "pod" are likely the correct spellings, since it's a proper English word, not an abbreviation, like "Plain Old Data". I mention this, so that we avoid having someone in ALICE read our documentation and then learn to read it "pee-o-dee".

docs/kubernetes_ecs.md
```diff
@@ -0,0 +1,70 @@
...
+* `get` => `kubectl get -f manifest.yaml -o jsonpath='{.spec.state}'` - queries exact field of resource (`state` in the example) inside cluster.
+
+These four commands allow us to deploy and monitor status of the deployed
+resource without necessity to interact with it directly. However `KubectlTask`
+expects that resource is the CRD [Task](/control-operator/api/v1alpha1/task_types.go).
+
+In order to activate `KubectlTask` you need to change yaml template
+inside the `ControlWorkflows` directory. Namely:
+
+* add path to the kubectl manifest as the first argument in `.command.arguments` field
```
knopers8, line 69: Huh, that explains my question from the other PR. Couldn't we just have a separate field?

core/task/taskclass/class.go
justonedev1: This is probably my misunderstanding, as I thought that it is a bug to implicitly change hook to direct.. especially when we have hooktask that is created only if `controlmode.HOOK` is present.
[see](https://github.com/AliceO2Group/Control/blob/master/executor/executable/task.go#L151-L164)

docs/kubernetes_ecs.md
justonedev1: yes, it would be cleaner... but I didn't want to make changes to existing templates just for POC that will be probably obsoleted quickly, as I didn't want to think what would need to change for this to apply

docs/kubernetes_ecs.md
justonedev1: fair enough

docs/kubernetes_ecs.md
knopers8: fair enough

docs/kubernetes_ecs.md
```diff
@@ -0,0 +1,70 @@
...
+* change `.control.mode` to either `kubectl_direct` or `kubectl_fairmq`
+You can find working template inside `control-operator/ecs-manifests/control-workflows/*_kube.yaml`
```
knopers8:
```suggestion
You can find working template inside `control-operator/ecs-manifests/control-workflows/*-kube.yaml`
```

executor/executable/kubectltask.go
```diff
@@ -0,0 +1,395 @@
...
+
+	command.Stdout = io.MultiWriter(os.Stdout, &stdoutBuf)
+	command.Stderr = io.MultiWriter(os.Stderr, &stderrBuf)
+
+	err = command.Run()
+	if err != nil {
+		log.WithFields(logrus.Fields{
+			"controlmode": task.Tci.ControlMode,
+			"name":        task.ti.Name,
+		}).WithError(err).Errorf("kubectl apply failed stderr: %s , stdin: %s", stderrBuf.String(), stdoutBuf.String())
```
knopers8:
```suggestion
		}).WithError(err).Errorf("kubectl apply failed stderr: %s , stdout: %s", stderrBuf.String(), stdoutBuf.String())
```

executor/executable/kubectltask.go
```diff
@@ -0,0 +1,395 @@
...
+		os.Setenv("FLP_UID", strconv.FormatInt(uid, 10))
+		os.Setenv("FLP_GID", strconv.FormatInt(gid, 10))
+
+		var strIds []string
+		for _, id := range supplementalIds {
+			strIds = append(strIds, strconv.FormatInt(id, 10))
+		}
+		supplementalString := "[" + strings.Join(strIds, ", ") + "]"
+
+		os.Setenv("FLP_SUPPLEMENTAL_GORUPS", supplementalString)
```
knopers8:
```suggestion
		os.Setenv("FLP_SUPPLEMENTAL_GROUPS", supplementalString)
```
typo?

docs/kubernetes_ecs.md
```diff
@@ -0,0 +1,70 @@
+# ECS with Kubernetes
+
+> ⚠️ **Warning**
+> All Kubernetes work done is in a stage of prototype.
+
+## Kubernetes Cluster
```
knopers8, line 6: My notes on what was generally missing here (or I just didn't see it):
- `make install` to register the task CRD
- make sure executor has access to a `~/.kube/config`
- kubectl does not react well to stfsender env var `http_proxy=""`, which is surrounded by additional quotes in its manifest, making `kubectl apply` fail. Easy fix - `http_proxy=`.

knopers8: ah, and one has to remember to create an image pull secret for the controller after `make deploy`, because only then the namespace already exists.

justonedev1: I think that some of those of those comments belong into `kubernetse_ecs.md`, but to the docs in control-operator. But those are fair comments. The reason why I didn't add them at the time of creation was that I didn't encounter those bcs k3s behaved a bit differently (can automatically read `~/.docker/config.json` and other things) But I will add those comments into the `control-operator` with maybe a link pointing from `kubernetes_ecs.md` there.

justonedev1: I don't know what you mean by `http_proxy` comment. Can you tell me what exactly should I add and what you did, I don't think that I had to do anything about that.

knopers8: > I think that some of those of those comments belong into ...

Yes, very likely, at some point I became a bit lost which PR is which :D

> I don't know what you mean by http_proxy comment. Can you tell me what exactly should I add and what you did, I don't think that I had to do anything about that.

I mean this line: https://github.com/AliceO2Group/ControlWorkflows/blob/15b05c6ea90cf61b322a921c9640300ac981ff4e/tasks/stfsender.yaml#L28

These env vars are somehow passed to the manifest and causing some quoting/escaping issue, I don't remember exactly.

core/task/taskclass/class.go
knopers8: then perhaps extract the bugfix to a separate PR and we can validate it.

executor/executable/task.go
```diff
@@ -90,6 +92,7 @@ func NewTask(taskInfo mesos.TaskInfo, sendStatusFunc SendStatusFunc, sendDeviceE
 
 	log.WithField("json", string(tciData[:])).
 		Trace("received TaskCommandInfo")
+	log.WithField("findme", "here").Info(string(tciData))
```
knopers8: maybe make it Info/devel at least and don't use "findme" "here" :D

core/task/taskclass/class.go
justonedev1: https://github.com/AliceO2Group/Control/pull/815

knopers8 (approved):
thank you!

PR #810, opened 2026-04-23
[OCTRL-1092] Improve metrics error reporting

justonedev1:


core/environment/transition_startactivity.go
```diff
@@ -72,6 +73,9 @@ func (t StartActivityTransition) do(env *Environment) (err error) {
 		return errors.New("cannot transition in NIL environment")
 	}
 
+	metric := transitionMetric("startactivity", env)
```
knopers8: Any reason not to use the same formatting as for the "official" transition names, i.e. `START_ACTIVITY`, `GO_ERROR`, etc?
```suggestion
	metric := transitionMetric(t.baseTransition.name, env)
```

justonedev1: changed

knopers8 (approved):
thank you!

PR #814, opened 2026-04-28
[OCTRL-1092] Add metrics to TryTransition

justonedev1:


common/monitoring/metric.go
```diff
@@ -78,6 +78,14 @@ const (
 	TIMEOUT   = "timeout"
 )
 
+func (metric *Metric) AddErr(err error) {
```
knopers8:
```suggestion
func (metric *Metric) AddResult(err error) {
```
How about that, given that it modifies the metric even when error is nil?

justonedev1: you cannot do this in golang as there already exists AddResult. Function overloading isn't allowed in go.

knopers8: aah, right, i completely missed that.

knopers8 (approved):
thank you

PR #815, opened 2026-05-08
Hook control mode was changed to direct in task

justonedev1:
During my kubernetes work, I found this weird forking of control mode in task.go which silently move hook to direct control task. This seems weird especially when we have dedicated HookTask ([see](https://github.com/AliceO2Group/Control/blob/master/executor/executable/task.go#L151-L164)). I cannot say whether this was done deliberately or whether it is a bug, but it feels like a bug.


knopers8 (approved):
I had a look around and it seems that this function is not called anywhere, maybe it was something added for `walnut`, judging by the author.

Seems safe to merge.

PR #816, opened 2026-05-11
[OCTRL-1088] Environment Controller

justonedev1:
Apart from environment controller I also created separate binaries for environment manager and task manager, which is the cause of most of the yaml piping.


control-operator/internal/controller/fairmq_transition.go
```diff
@@ -29,7 +29,7 @@ import (
 	"fmt"
 	"strings"
 
-	pb "github.com/AliceO2Group/ControlOperator/internal/controller/protos/generated"
+	pb "github.com/AliceO2Group/Control/operator/internal/controller/protos/generated"
```
knopers8, line 32: hmm, isn't it `control-operator`?
```suggestion
	pb "github.com/AliceO2Group/Control/control-operator/internal/controller/protos/generated"
```

control-operator/internal/controller/environment_controller.go
```diff
@@ -0,0 +1,311 @@
...
+				log.Error(fmt.Errorf("didn't find required env: %s", env), "failed to fill in env vars from template")
+				return &reconcile.Result{}, nil
+			}
+		}
+
+		task.Spec.Pod = *template.Spec.Pod.DeepCopy()
+		task.Spec.Control = *template.Spec.Control.DeepCopy()
+
+		if foundIdx := slices.IndexFunc(taskReference.Env, func(envVar v1.EnvVar) bool { return envVar.Name == "OCC_CONTROL_PORT" }); foundIdx == -1 {
+			log.Error(fmt.Errorf("didn't find OCC_CONTROL_PORT in env"), "failed to fill in env vars from template")
```
knopers8, line 89: Shouldn't we return if there is error?

knopers8 (dismissed):
Thanks, I took some time to get acquainted with the news and finally tried running it today - it works!

I have some minor questions, but it can be merged already. Please put a human-readable title to the PR though.

control-operator/internal/controller/fairmq_transition.go
justonedev1: you are correct, I will address it in separate PR as proper renaming would touch multiple files and I can do at least once something properly in it's own PR/commit :D

control-operator/internal/controller/environment_controller.go
justonedev1: yes, to have symetrical behavior with the other error handling in the controller... However I am not sure that it is the correct behavior here, as it drops the deployment, which might not be desired outcome. I added TODO into the code to address this potentialy

PR #820, opened 2026-05-26
Rename operator

justonedev1:
renaming package operator -> control-operator so the structure is according to the name of folder structure, which is control-operator


knopers8 (approved):
thank you

PR #821, opened 2026-05-27
[OCTRL-1090] Create k8s client to allow deployment of custom CRDs programmatically

justonedev1:


control-operator/pkg/client/client.go
```diff
@@ -0,0 +1,133 @@
...
+
+func buildConfig(kubeconfigPath string) (*rest.Config, error) {
+	if kubeconfigPath != "" {
+		return clientcmd.BuildConfigFromFlags("", kubeconfigPath)
+	}
+	if config, err := rest.InClusterConfig(); err == nil {
+		return config, nil
+	}
+
+	return nil, fmt.Errorf("ECS isn't running in a container and you didn't pass any kubernetes config")
```
knopers8: maybe it's worth to print also `err`?

control-operator/pkg/client/suite_test.go
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * In applying this license CERN does not waive the privileges and
+ * immunities granted to it by virtue of its status as an
+ * Intergovernmental Organization or submit itself to any jurisdiction.
+ */
+
+package client_test
```
knopers8, line 25: not for this PR, but I would propose that the tests in `control-operator` are added to CI.

knopers8 (dismissed):

control-operator/pkg/client/client.go
justonedev1: I assume that logging is done by the user of the client. It might not be wanted behavior to log error multiple times (caller, and the client itself)

knopers8: But `err` received from `rest.InClusterConfig` is swallowed by this function, it's not propagated to the user. One could either print it or include it in the newly created error.

justonedev1: fair enough, I missed that when refactoring

knopers8 (approved):

PR #822, opened 2026-06-25
OCTRL-1095: Create observability stack for kubernetes operators

justonedev1:


control-operator/api/v1alpha1/task_types.go
```diff
@@ -88,13 +88,24 @@ type TaskSpec struct {
 	NodeName string `json:"nodeName,omitempty"`
 }
 
+const (
+	ConditionPodReady          = "PodReady"
+	ConditionGRPCConnected     = "GRPCConnected"
+	ConditionStateInitialized  = "StateInitialized"
+	ConditionStateTransitioned = "StateTransitioned"
```
knopers8: What do they mean exaclty? State transition started and finished?

control-operator/internal/controller/task_controller.go
```diff
@@ -385,6 +406,22 @@ func (r *TaskReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
+func (r *TaskReconciler) recordCondition(ctx context.Context, t *aliecsv1alpha1.Task, condType string, condStatus metav1.ConditionStatus, reason, message string) error {
+	patch := client.MergeFrom(t.DeepCopy())
+	eventType := v1.EventTypeNormal
+	if condStatus == metav1.ConditionFalse {
```
knopers8, line 412: out of curiosity, is a false condition by convention something bad, given that we automatically make it a warning?

docs/k3s_observability.md
```diff
@@ -0,0 +1,124 @@
...
+| ALIECS CRDs (Task, Environment, TaskTemplate) | `RequestResponse` (full spec) | create, update, patch, delete |
+| Pods | `Metadata` (no body) | create, delete |
+
+`RequestResponse` means the full request and response body is logged — i.e. the complete spec of every Task and Environment CRD at the time it was created or modified. This gives a persistent record of what was deployed even after the CRD is deleted.
+
+`managedFields` is stripped at source via `omitManagedFields: true` in the audit policy. This field uses `.` as a JSON key (Kubernetes FieldsV1 format), which OpenSearch rejects. Removing it at the kube-apiserver level is cleaner than filtering it in the pipeline.
+
+Pod deletion (which sets the pod to Terminating) is captured at `Metadata` level via `verb: delete`.
+
+What is **not** captured: pod status transitions (Running → Terminating → Succeeded/Failed) — these are `patch` operations on the Pod object and are excluded to avoid noise.
```
knopers8, line 118: hmm, why noise? i imagine those could be helpful as well.

justonedev1: noise in a sense that they should be already inside the Pod object, so it isn't necessary to copy them (at least that was the reasoning)

control-operator/api/v1alpha1/task_types.go
justonedev1: ConditionStateInitialized means that we were able to query the state of the OCC task
ConditionStateTransitioned just reports the Transitions between states

control-operator/internal/controller/task_controller.go
justonedev1: How I think about this: All conditions are supposed to be true for a healthy Task, so when something isn't true (eg. we failed to connect via grpc) I see it as something abnormal, so I issue a warning... maybe error event should be triggered.

justonedev1: but from the POV of controller it isn't good or bad, it just reflects the actual state. I wanted different type of even so it could be easily filtered through.

knopers8: ok, i was just wondering how future-proof it is to assume that false conditions always deserve a warning/error. anyway, this should be easy to refactor.

control-operator/api/v1alpha1/task_types.go
knopers8: hmm, then maybe i'd propose to rename them to have less ambiguous meaning:
- StateAccessible
- LastStateTransitionResult

knopers8: (subject to discussion of course)

control-operator/internal/controller/task_controller.go
justonedev1: Yes, there should be move from POC stage to alpha/beta stage where we have consistent and agreed upon abnormal behavior handling (I mean any behavior out of ordinary healthy Task/Environment)

knopers8 (approved):

PR #824, opened 2026-07-03
Run control-operator tests in CI

knopers8:
Just checking if it works.

### DIFF
--- a/.github/workflows/control.yml
+++ b/.github/workflows/control.yml
@@ -23,4 +23,5 @@ jobs:
       - run: (make vendor)
       - run: (make)
       - run: (make test)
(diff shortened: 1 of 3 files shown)


justonedev1 (dismissed):

.github/workflows/control.yml
```diff
@@ -23,3 +23,4 @@ jobs:
       - run: (make vendor)
       - run: (make)
       - run: (make test)
+      - run: (cd control-operator && make test)
```
justonedev1: maybe add just as well simple `make` as well to check if we can build everything?

knopers8: let's try. now it can't find protoc anyway.

PR #827, opened 2026-07-03
[control-operator] Environment CRD finalizer

justonedev1:
this finalizer is used to stop Environment CRD deletion until all connected tasks are gone


knopers8 (approved):
thank you!

PR #830, opened 2026-07-03
[control-operator] finalizer blocks deletion of task until all pods are gone

justonedev1:
can be merged, if you think that it is good enough

### DIFF
--- a/control-operator/internal/controller/task_controller.go
+++ b/control-operator/internal/controller/task_controller.go
@@ -283,13 +283,31 @@ func (r *TaskReconciler) handleFinalizer(ctx context.Context, t *aliecsv1alpha1.
 		}
 	} else {
 		if controllerutil.ContainsFinalizer(t, taskFinalizer) {
-			log.Info("Cleaning up gRPC connection before deletion")
+			log.Info("Finalizer found, starting cleanup")
 			if client, exists := clientsForContainers[t.Name]; exists {
+				log.Info("Cleaning up gRPC connection before deletion")
 				if err := client.Close(); err != nil {
 					log.Error(err, "Failed to close gRPC client during deletion")
 				}
 				delete(clientsForContainers, t.Name)
+				log.Info("gRPC cleaned")
 			}
+
+			pod := &v1.Pod{}
+			err := r.Get(ctx, types.NamespacedName{Name: podNameFromTask(t.Name), Namespace: t.Namespace}, pod)
+			if err == nil {
+				if pod.DeletionTimestamp.IsZero() {
+					log.Info("Deleting pod before removing finalizer")
+					if err := r.Delete(ctx, pod); err != nil && !errors.IsNotFound(err) {
+						return ctrl.Result{}, true, err
+					}
+				}
+				log.Info("Waiting for pod to terminate before removing finalizer")
+				return ctrl.Result{}, true, nil
+			} else if !errors.IsNotFound(err) {
+				return ctrl.Result{}, true, err
+			}
+			log.Info("POD cleaned")
 
 			controllerutil.RemoveFinalizer(t, taskFinalizer)
 			if err := r.Update(ctx, t); err != nil {
--- a/control-operator/internal/controller/task_controller.go
+++ b/control-operator/internal/controller/task_controller.go
@@ -270,53 +270,65 @@ func (r *TaskReconciler) consumeGRPCConsumerIfReady(ctx context.Context, t *alie
 	return ctrl.Result{}
 }
 
+// handleFinalizer sets the finalizer, when Task isn't marked for deletion, and when deletion reconcile is triggered
+// it then clears the gRPC connection and deletes the underlying POD and doesn't allow Task to be deleted
+// until POD is gone.
 // Note: if you add finalizer to a task you cannot delete it unless you remove the finalizer, run:
 // kubectl patch task --type=json -p='[{"op": "remove", "path": "/metadata/finalizers"}]'
 func (r *TaskReconciler) handleFinalizer(ctx context.Context, t *aliecsv1alpha1.Task, log logr.Logger) (ctrl.Result, bool, error) {
 	if t.DeletionTimestamp.IsZero() {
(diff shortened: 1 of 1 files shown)


knopers8 (dismissed):
The logic seems OK, but please try to simplify the logic and reduce nesting similarly to environment's handleFinalizer. Perhaps in a separate PR, but as you prefer.

control-operator/internal/controller/task_controller.go
```diff
@@ -270,35 +270,65 @@ func (r *TaskReconciler) consumeGRPCConsumerIfReady(ctx context.Context, t *alie
...
+		return ctrl.Result{}, true, err
+	}
+
+	controllerutil.RemoveFinalizer(t, taskFinalizer)
+	return ctrl.Result{}, true, r.Update(ctx, t)
+}
+
+// podDeleted deletes the Task's Pod if it still exists, and reports whether it is
+// fully gone yet so the caller can wait for termination before removing the finalizer.
+func (r *TaskReconciler) podDeleted(ctx context.Context, t *aliecsv1alpha1.Task, log logr.Logger) (bool, error) {
```
knopers8: The name of the function is misleading, as it does not indicate that it attempts to delete the pod. Sure, the behaviour is explained better in a comment, but one can only notice it when navigating to the function implementation. Please rename to sth like "deletePod" and also name the return value in the declaration.
```suggestion
func (r *TaskReconciler) deletePod(ctx context.Context, t *aliecsv1alpha1.Task, log logr.Logger) (podGone bool, error) {
```

However, given that you have to negate the returned value after calling `podDeleted`, you could even return `podExists` for (arguably) easier reading?

control-operator/internal/controller/task_controller.go
```diff
@@ -270,35 +270,65 @@ func (r *TaskReconciler) consumeGRPCConsumerIfReady(ctx context.Context, t *alie
 	return ctrl.Result{}
 }
 
+// handleFinalizer sets the finalizer, when Task isn't marked for deletion, and when deletion reconcile is triggered
+// it then clears the gRPC connection and deletes the underlying POD and doesn't allow Task to be deleted
+// until POD is gone.
 // Note: if you add finalizer to a task you cannot delete it unless you remove the finalizer, run:
 // kubectl patch task --type=json -p='[{"op": "remove", "path": "/metadata/finalizers"}]'
 func (r *TaskReconciler) handleFinalizer(ctx context.Context, t *aliecsv1alpha1.Task, log logr.Logger) (ctrl.Result, bool, error) {
```
knopers8: please document what the function returns, especially what the bool means.

control-operator/internal/controller/task_controller.go
```diff
@@ -270,35 +270,65 @@ func (r *TaskReconciler) consumeGRPCConsumerIfReady(ctx context.Context, t *alie
...
+	}
 
-			controllerutil.RemoveFinalizer(t, taskFinalizer)
-			if err := r.Update(ctx, t); err != nil {
-				return ctrl.Result{}, true, err
-			}
-		}
+	if !controllerutil.ContainsFinalizer(t, taskFinalizer) {
 		return ctrl.Result{}, true, nil
 	}
```
knopers8, line 288: Which situation could lead us to this case, i.e. scheduled for deletion, but no finalizer? Given that we ask `Reconcile` to return, won't it result in an infinite loop?

control-operator/internal/controller/task_controller.go
```diff
@@ -270,35 +270,65 @@ func (r *TaskReconciler) consumeGRPCConsumerIfReady(ctx context.Context, t *alie
...
-	return ctrl.Result{}, false, nil
+
+	log.Info("Finalizer found, starting cleanup")
```
knopers8, line 290:
```suggestion
	log.Info("Deletion timestamp exists and finalizer found, starting cleanup")
```
I would propose to be more clear about conditions for starting cleanup, so one is not misled into thinking that it's enough to have finalizer present to start cleanup.

knopers8 (changes requested):
Thanks for the refactoring, it does look clearer! I would still have a few suggestions and questions though.

control-operator/internal/controller/task_controller.go
justonedev1: The timestamp is automatically added by k8s, not by us when there is a request to delete any resource, so in my opinion it isn't necessary to repeat it.

control-operator/internal/controller/task_controller.go
justonedev1: It would not result in infinite loop because it can happen only when the object was finalised but not deleted, aka in a process of deletion. Basically indempotency check. From the point of view of this controller this object is done/gone and shouldn't be acted up. But Reconciliation can be triggered by k8s and this causes Reconcile loop for given object to be no-op as it should be

control-operator/internal/controller/task_controller.go
justonedev1: agreed on the bad naming of the function. I don't really agree about renaming and inverting the return value, as I always assumed that the bool result coming from the function means whether the function did it's job. However in golang this is signified more by the `error` return value. So if you really want, I can rename and invert the return value.

control-operator/internal/controller/task_controller.go
knopers8: Thanks. To rephrase and confirm my understanding: this case may happen when the corresponding Pod is gone, the Task finalizer was removed, but for whatever reason the Task object still lives in etcd (e.g. has another finalizer?). We assume our `Reconcile` is not responsible for any further actions.

Let me know if I'm still missing it though :D

control-operator/internal/controller/task_controller.go
knopers8: > as I always assumed that the bool result coming from the function means whether the function did it's job

That makes sense, all good then.

control-operator/internal/controller/task_controller.go
knopers8: ok, fair enough

knopers8 (approved):

PR #831, opened 2026-07-03
[control-operator] disable grpc streams for time being

justonedev1:
This is nice TODO to fix grpc OCC streams to be useful for task controller, which they aren't now. I just annotated with remark in the code.

### DIFF
--- a/control-operator/internal/controller/grpc_client.go
+++ b/control-operator/internal/controller/grpc_client.go
@@ -171,11 +171,16 @@ func (c *OccClient) ConsumeIfReady(ctx context.Context) bool {
 		c.log.V(1).Info("connection is in different state than ready", "conn state", connState.String())
 		return false
 	}
-	clientCtx, clientCancel := context.WithCancel(context.Background())
+	// clientCtx, clientCancel := context.WithCancel(context.Background())
+	_, clientCancel := context.WithCancel(context.Background())
 	c.cancel = &clientCancel
 
-	go c.ConsumeEventStream(clientCtx)
-	go c.ConsumeStateStream(clientCtx)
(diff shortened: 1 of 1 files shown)


control-operator/internal/controller/grpc_client.go
```diff
@@ -171,11 +171,16 @@ func (c *OccClient) ConsumeIfReady(ctx context.Context) bool {
...
+	// clientCtx, clientCancel := context.WithCancel(context.Background())
+	_, clientCancel := context.WithCancel(context.Background())
 	c.cancel = &clientCancel
 
-	go c.ConsumeEventStream(clientCtx)
-	go c.ConsumeStateStream(clientCtx)
+	// TODO: Streams are not used anywhere now so we turn them off until they are fixed/useful
+	// 		 they are not useful for now because any Transition call imediately returns final
+	//		 status. So for now we don't need them. Of course the problem is that if something
+	//		 changes inside OCC container we have no way how to detect it now.
```
knopers8, line 181: I don't quite understand what is broken at the moment. Don't we want to log spontaneous changes in the controlled task, even if we don't react to them?

justonedev1: we can log those, but the reason why I disabled them was that this actually throws errors, so I threw away this whole thing, with an asterisk to rething and redo it:
https://github.com/AliceO2Group/Control/blob/9442b814ea7b08235f3af9fef0c0a9ea814dc699/control-operator/internal/controller/grpc_client.go#L305

It was also the reason why I made this as a `TODO` and didn't delete it at all. I can make a Jira ticket about it as well if you want.

knopers8: indeed, please create a ticket, thanks!

knopers8 (approved):

PR #833, opened 2026-07-08
[OCTRL-1091] Integrate k8s client into ECS core

justonedev1:
This is actual implementation of the ECS Kubernetes bridge. 

Before merging it needs the [OCTRL-1090](https://github.com/AliceO2Group/Control/pull/821)

The goal of this PR was to take existing code that deploys Mesos tasks and adjusts it for the kubernetes client usage, basically linking CRDs defined and managed by managers in control-operator to the existing ECS task infrastructure and classes. The code from managerk8s.go was inspired roughly accroding to this table (generated by llm, but I went through it):

| `managerk8s.go` | Mesos counterpart | Location |
|---|---|---|
| `createK8sTaskEntries` — InboundChannels loop, IPC endpoint, `::global` alias | `makeTaskForMesosResources` — same loop, IPC/TCP handling, global alias | `scheduler.go:1346` |
| `buildK8sNodeTaskRefs` — `O2_ROLE`/`O2_SYSTEM` env var injection loop | `makeTaskForMesosResources` — `fillEnvDefault` loop for the same two vars | `scheduler.go:1458` |
| `buildK8sNodeTaskRefs` — ArgsCLI trimming/filtering | `makeTaskForMesosResources` — `cmd.Arguments` handling | `scheduler.go:1437` |
| `newTaskForKubernetes` | `newTaskForMesosOffer` | `manager.go:169` |
| `deployKubernetesTasks` — overall flow | `acquireTasks` — Mesos deployment block | `manager.go:550` |
| `configureK8sTasks` — `BuildPropertyMap` per task, global bind map | `configureTasks` — bind map building + `BuildPropertyMaps` | `manager.go:804` |
| `transitionAndWaitK8sEnvState` — send transition, block until state reached | `transitionTasks` — send Mesos command, block on response | `manager.go:901` |
| `killK8sTask` | `doKillTask` | `manager.go:1238` |
| `killK8sEnvironment` | `releaseTasks` | `manager.go:733` |
| `process` — route watch event to roster task (called from `watchK8sTasks`) | `updateTaskState` / `updateTaskStatus` | `manager.go:1071` |
| `watchK8sTasks` — retry loop on watcher failure | Mesos subscribe retry loop | `scheduler.go` |
| `nodeNameFromDescriptor` — scan constraints for `machine_id` | Mesos offer attribute matching / `Satisfy(targetConstraints)` | `scheduler.go:743` |
| `k8sEnvRegistry`, `createK8sEnvironmentCRD`, `waitForK8sEnvState`, `cmdEnvToK8sEnvVars` | *(no direct Mesos equivalent)* | — |

It contains multiple changes:
- upgrading deps after rebasing on master
- new `controlmode`s to distinguish between Kubernetes and Mesos tasks
- k8s manager which translates existing mesos infrastructure to kubernetes (creation, updating, transitioning ,...)
- small change for environment CRDs
- control-workflow manifests for `readout`, `stfbuilder-senderoutput` and `stfsender`
- simple documentation


core/task/manager.go
```diff
@@ -144,6 +149,13 @@ func NewManager(shutdown func(), internalEventCh chan<- event.Event) (taskman *M
 	taskman.tasksToDeploy = taskman.schedulerState.tasksToDeploy
 	taskman.reviveOffersTrg = taskman.schedulerState.reviveOffersTrg
 	taskman.ackKilledTasks = safeacks.NewAcks()
+	taskman.k8sEnvs = newK8sEnvRegistry()
+	if k8sClient, k8sErr := newK8sClientFromViper(); k8sErr != nil {
+		log.WithError(k8sErr).Warn("K8s client init failed, K8s tasks disabled")
```
knopers8: perhaps warn/devel for the time being, so no one panics when using AliECS in a "classic" setup?

docs/kubernetes_ecs.md
```diff
@@ -23,14 +24,41 @@ is ability to run Pods with privileged permissions and also under user `flp`.
...
+
+The bridge works as follows:
+
+1. **ECS core** (running outside the cluster) uses a typed Kubernetes client (`control-operator/pkg/client`) backed by `controller-runtime` to interact with the cluster.
+2. When ECS deploys a task whose control mode is `kubernetes_direct` or `kubernetes_fairmq`, it creates a **Task CRD** and, if needed, an **Environment CRD** in the cluster instead of submitting a Mesos offer.
+3. Inside the cluster, the **task-manager** and **environment-manager** operators (see `control-operator/cmd/`) watch these CRDs and reconcile the desired state — scheduling Pods, driving OCC gRPC state transitions, and writing status back into the CRD.
+4. ECS watches the CRDs via the Kubernetes Watch API and reacts to status changes exactly as it would to Mesos task updates, keeping the rest of the ECS state machine intact.
+
+This design lets ECS reuse its existing environment lifecycle, configuration generation, and monitoring logic while outsourcing Pod scheduling and OCC communication to the in-cluster operators. The relevant code can be found in manager.go and managerk8s.go.
```
knopers8: We might need to be more specific, as there is `task/manager.go` and `environment/manager.go`

core/task/manager.go
```diff
@@ -476,6 +488,33 @@ func (m *Manager) acquireTasks(envId uid.ID, taskDescriptors Descriptors) (err e
 		tasksToRun = append(tasksToRun, taskDescriptors...)
 	}
 
+	// Split K8s descriptors out before the Mesos retry loop.
+	var k8sDescriptors Descriptors
+	{
+		mesos := make(Descriptors, 0, len(tasksToRun))
```
knopers8:
```suggestion
		mesosDescriptors := make(Descriptors, 0, len(tasksToRun))
```

core/task/managerk8s.go
```diff
@@ -0,0 +1,452 @@
...
+	if err != nil {
+		log.WithField("partition", envId).WithError(err).Error("failed to build K8s node task refs")
+		return nil, err
+	}
+
+	for _, e := range entries {
+		m.roster.append(e.task)
+	}
+
+	envCRDName, err := m.createK8sEnvironmentCRD(ctx, envId, nodeToRefs)
```
knopers8: I might be mistaken, but isn't it a custom object that we are creating, not Custom Resource Definition? My understanding is that a CRD defines what fields our custom objects should have at the new custom resource/endpoint.
```suggestion
	envCRDName, err := m.createK8sEnvironmentObject(ctx, envId, nodeToRefs)
```
So if I am right, any occurence of "CRD" should be replaced here with "Object" or removed.

core/task/managerk8s.go
```diff
@@ -0,0 +1,452 @@
...
+				return fmt.Errorf("watch channel closed waiting for Environment %s", envCRDName)
+			}
+			env, ok := ev.Object.(*v1alpha1.Environment)
+			if !ok {
+				continue
+			}
+			log.WithField("partition", envCRDName).
+				WithField("crd", envCRDName).Debug("received some environment update")
+			if env.Name == envCRDName {
+				log.WithField("crd", envCRDName).
```
knopers8:
```suggestion
				log.WithField("crd", envCRDName).
					WithField("partition", envCRDName).
```

knopers8 (changes requested):
Thanks, i have comments mostly to naming, while the logic and approach seems OK for me. 

task/manager.go becomes less clean due to placing kubernetes stuff there, but I wouldn't worry too much with creating a common abstraction for k8s + mesos, since this code is subject to further rewriting or starting it from scratch without mesos in the first place.

core/task/managerk8s.go
justonedev1: no this is actual CRD that we are creating and filling in

knopers8: Well, doesn't that happen during `make install`? Why do we have to do it at environment creation?

https://github.com/AliceO2Group/Control/blob/763366e25dd0841d93a650c117fa2c22daf1ac4d/control-operator/Makefile#L202

justonedev1: Ah, I understand now. No I am not installing anything, cluster alreay needs to know about this CRD to create an instance of it (that is what install is for). But this code does the same as what you do if you use `kubectl apply -f environment.yaml`. It just describes the instance of environment by filling in the instance of Environment CRD with required tasks and other properties and applies it to the cluster.

knopers8: Exactly. So my point was that Environment CRD is what we create by `kubectl apply` during `make install`. As in the [doc](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#customresourcedefinitions):

> Defining a CRD object creates a new custom resource with a name and schema that you specify.

Then:

> A resource is an endpoint in the [Kubernetes API](https://kubernetes.io/docs/concepts/overview/kubernetes-api/) that stores a collection of [API objects](https://kubernetes.io/docs/concepts/overview/working-with-objects/#kubernetes-objects) of a certain kind; for example, the built-in pods resource contains a collection of Pod objects.

So what we are creating here is an object (instance) of the Environment custom resource.

justonedev1: yes, that is what I am saying and the reason for the name. If you use just "Object" it is misleading as in "what object" it is concretly defined as instance of CRD... I can rename it to createK8sEnvironmentCRDInstance, but I think that leaving CRD at all from the name is obscuring information from reader

justonedev1: I changed naming to what we agreed in person

knopers8 (approved):
thank you

PR #848, opened 2026-08-12
test deploy github workflow

justonedev1:
I created a workflow that tests, creates and pushes images into github repository. Reacting on git tag in a shape `control-operator-v*` where version must match with VERSION variable in Makefile in control-operator directory.
It was already run and results are here:
https://github.com/orgs/AliceO2Group/packages

### DIFF
--- a/.github/workflows/control-operator-release.yml
+++ b/.github/workflows/control-operator-release.yml
@@ -1,0 +1,59 @@
+name: AliECS Control Operator Release
+
+on:
+  push:
+    tags:
+      - 'control-operator-v*'
+
+permissions:
+  contents: read
+  packages: write
+
+defaults:
+  run:
+    working-directory: control-operator
+
+jobs:
+  release:
+    name: Test, Build & Publish Images
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: actions/setup-go@v7
+        with:
+          go-version: '1.25.0'
+
+      - uses: arduino/setup-protoc@v3
+        with:
+          version: '35.1'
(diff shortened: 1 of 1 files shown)


.github/workflows/control-operator-release.yml
```diff
@@ -0,0 +1,59 @@
...
+jobs:
+  release:
+    name: Test, Build & Publish Images
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: actions/setup-go@v7
+        with:
+          go-version: '1.26.5'
```
knopers8: maybe let's use a consistent version with https://github.com/AliceO2Group/Control/blob/master/control-operator/go.mod (that can also mean bumping it there)

.github/workflows/control-operator-release.yml
```diff
@@ -0,0 +1,59 @@
...
+
+      - uses: arduino/setup-protoc@v3
+        with:
+          version: '35.1'
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
knopers8, line 30: why do we need it? don't we get protoc anyway through the recently fixed control-operator/Makefile?

.github/workflows/control-operator-release.yml
justonedev1: I set it to 1.25.0 same as go.mod

.github/workflows/control-operator-release.yml
knopers8: explained in person - we don't install protoc in `tools/`, just the go specific tooling

knopers8 (approved):
thank you!

PR #849, opened 2026-08-20
Docs for creation lab K8s

justonedev1:


docs/lab-k8s-install.md
```diff
@@ -0,0 +1,253 @@
...
+It is possible to install kubeadm/cluster with SE enabled, but for now I chose to follow guide to the t.
+
+```
+sudo setenforce 0
+sudo sed -i 's/^SELINUX=enforcing$/SELINUX=permissive/' /etc/selinux/config
+```
+
+- Verify all required kernel modules are loaded
+
+RHEL 10 can have problems with having all required kernel modules, you can check:
```
knopers8, line 23: I'm not sure where this comes from, can't find corresponding advice in the k8s doc and I didn't have to do any of this when installing. When migrating to an ansible playbook, I would try first without.

Maybe it comes from Calico?

docs/lab-k8s-install.md
```diff
@@ -0,0 +1,253 @@
...
+```
+enable the systemd service
+```
+systemctl daemon-reload
+systemctl enable --now containerd
+```
+RHEL 10 is running cgroups v2 as a default, so we need containerd to use systemd as cgroups driver so add following to the
+`/etc/containerd/config.toml` together with some other changes (like bin dir and other), contents of the config.toml are:
+```
+cat /etc/containerd/config.toml
```
knopers8, line 108: i'd use v3, probably more future-proof:
```
version = 3

[plugins]
  [plugins."io.containerd.cri.v1.images"]
    pinned_images = { sandbox = "registry.k8s.io/pause:3.10" }
  [plugins."io.containerd.cri.v1.runtime"]
    [plugins."io.containerd.cri.v1.runtime".cni]
      bin_dir = "/usr/libexec/cni/"
      conf_dir = "/etc/cni/net.d"
    [plugins."io.containerd.cri.v1.runtime".containerd.runtimes.runc]
      [plugins."io.containerd.cri.v1.runtime".containerd.runtimes.runc.options]
        SystemdCgroup = true
  [plugins."io.containerd.internal.v1.opt"]
    path = "/var/lib/containerd/opt"
```

docs/lab-k8s-install.md
```diff
@@ -0,0 +1,253 @@
...
+kubeadm token create --print-join-command
+```
+this command will give you the exact command you are supposed to run on on worker node and it whould look like
+```
+(only demonstrative example) kubeadm join ip.addr:port ...
+```
+
+
+# To init Control plane node
+- Open following ports: [k8s](https://kubernetes.io/docs/reference/networking/ports-and-protocols/), [calico](https://docs.tigera.io/calico-enterprise/latest/operations/troubleshoot/troubleshooting)
```
knopers8, line 159: that page in calico does not list any ports

docs/lab-k8s-install.md
```diff
@@ -0,0 +1,253 @@
...
+after copying cluster config you can check that control-plane is responding (eg.):
+```
+kubectl get nodes
+NAME                          STATUS     ROLES           AGE   VERSION
+alio2-bld4-mx-k8s01.cern.ch   NotReady   control-plane   17m   v1.36.3
+```
+
+The reason why cluster is responding NotReady is that we have no network plugin.
+
+- Calico
```
knopers8, line 237: I think we could go with Cilium, since ATLAS have experience with both and picked it. They measured that Cilium scales much better with number of Pods (quadratic vs linear, IIRC). I would not diverge at this point unless we have a reason for it.

knopers8 (commented):
again, some comments for migrating this to ansible.

docs/lab-k8s-install.md
justonedev1: This was taken directly from k8s docs 🤷‍♂️

docs/lab-k8s-install.md
justonedev1: That I found it wrong and I am sorry for that...

docs/lab-k8s-install.md
knopers8: I remember getting a part of it from k8s docs, but then needing to extend it with some other snippet which was in version 3. Maybe they updated the documentation in the meantime and they put the complete thing...?

docs/lab-k8s-install.md
```diff
@@ -0,0 +1,253 @@
...
+---
+apiVersion: kubelet.config.k8s.io/v1beta1
+kind: KubeletConfiguration
+cgroupDriver: systemd
+```
+
+Explanation of the settings:
+`localAPIEndpoint`: VMware network has two NICs, one connected to gpn and other for internal communication. this IP is the internal one, which is not a default so it is better to specifically list it 
+`networking`: this defines the pod IP range, and it has to satisfy two things — (1) it must not overlap your existing networks (10.163.0.0/16 internal net or 128.141.175.0/24 gpn), and (2) it must exactly match the CIDR configured in Calico's own custom-resources.yaml (spec.calicoNetwork.ipPools[].cidr),
+`cgroupDriver`: kubelet will use systemd as it needs to match with containerd and systemd is recommended as cgroup driver for cgroups v2
```
knopers8, line 206: I'll probably leave that out, because systemd seems to be the default:

> In v1.22 and later, if the user does not set the cgroupDriver field under KubeletConfiguration, kubeadm defaults it to systemd.

docs/lab-k8s-install.md
knopers8: For the record, kube-proxy cannot modify iptables without these modules. Some more are needed also for cilium.

